PR #6, opened 2020-03-30
Fix type errors

wemeetagain:
A previous commit changed our `hash` implementation from `hash(Buffer) => Buffer` to `hash(Uint8Array) => Uint8Array`

This caused a few type errors that are cleaned up here.

### DIFF
--- a/src/util/merkleize.ts
+++ b/src/util/merkleize.ts
@@ -19,7 +19,7 @@ export function merkleize(chunks: Buffer[], padFor = 0): Buffer {
       chunks.push(zeroHashes[layer]);
     }
     for (let i = 0; i < chunks.length; i += 2) {
-      chunks[i / 2] = hash(chunks[i], chunks[i + 1]);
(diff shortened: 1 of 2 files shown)


tuyennhv:
oh what's the down side of it? I thought it's fine in term of memory usage:
```
/**
     * When passed a reference to the .buffer property of a TypedArray instance,
     * the newly created Buffer will share the same allocated memory as the TypedArray.
     * The optional {byteOffset} and {length} arguments specify a memory range
     * within the {arrayBuffer} that will be shared by the Buffer.
     *
     * @param arrayBuffer The .buffer property of any TypedArray or a new ArrayBuffer()
     */
    static from(arrayBuffer: ArrayBuffer | SharedArrayBuffer, byteOffset?: number, length?: number): Buffer;
    
```

mpetrunic:
You didn't pass ArrayBuffer, in current code, you are using 
this constructor: https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_array
which allocates new array in memory and it's doing that in a loop a lot of times

PR #14, opened 2020-04-22
Add isFooType functions

wemeetagain:
Adds the following exported functions:
- `isUintType`
- `isNumberUintType`
- `isBigIntUintType`
- `isBooleanType`
- `isListType`
- `isVectorType`
- `isBitListType`
- `isBitVectorType`
- `isByteVectorType`
- `isContainerType`
- `isRootType`

### DIFF
--- a/src/types/basic/uint.ts
+++ b/src/types/basic/uint.ts
@@ -24,7 +24,17 @@ export class UintType<T> extends BasicType<T> {
   }
 }
 
+export const NUMBER_UINT_TYPE = Symbol.for("ssz/NumberUintType");
+
+export function isNumberUintType(type: unknown): type is NumberUintType {
+  return isTypeOf(type, NUMBER_UINT_TYPE);
+}
+
 export class NumberUintType extends UintType<number> {
+  constructor(options: IUintOptions) {
+    super(options);
+    this._typeSymbols.add(NUMBER_UINT_TYPE);
+  }
   assertValidValue(value: unknown): asserts value is number {
     if (!(Number.isSafeInteger(value as number) || value === Infinity)) {
       throw new Error("Uint value is not a number");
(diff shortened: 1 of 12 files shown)


src/types/basic/uint.ts
```diff
@@ -1,15 +1,23 @@
...
   byteLength: number;
 }
 
+export const UINT_TYPE = Symbol.for("ssz/UintType");
+
+export function isUintType<T>(type: unknown): type is UintType<T> {
+  return isTypeOf(type, UINT_TYPE);
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
3xtr4t3rr3str14l, line 14: are you planning on working towards removing these eslint-disable's?

src/types/composite/list.ts
```diff
@@ -17,6 +18,13 @@ type ListTypeConstructor = {
   new<T extends List<any>>(options: IListOptions): ListType<T>;
 };
 
+export const LIST_TYPE = Symbol.for("ssz/ListType");
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function isListType<T extends List<any>=List<any>>(type: unknown): type is ListType<T> {
```
3xtr4t3rr3str14l, line 24: same question as above, but i see that the type `ListType` already uses this `List<any>` so i guess i'm also wondering if we can find a way around the any's there too.  would using `unknown` type be possible (such as `List<unknown>`?  or maybe some other way?

src/types/basic/uint.ts
wemeetagain: whoops, it looks like a few are superfluous

src/types/composite/list.ts
wemeetagain: I'm copying the pattern below (in this file). I think we can revisit these `any`s in list.ts and vector.ts, but should probably do it separately to this PR.

wemeetagain:
@mpetrunic :+1: will add an optional name in a future PR
I want to expose these symbols in an override of `toString` and if the name exists, we can use it there too.

PR #16, opened 2020-04-27
Fix far future deserialization

mpetrunic:
I realize we will also have issue because we serialize FAR_FUTURE_EPOCH as "Infinity" while other clients uses number defined in [spec](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#constants).

### DIFF
--- a/src/types/basic/uint.ts
+++ b/src/types/basic/uint.ts
@@ -18,6 +18,15 @@ export class UintType<T> extends BasicType<T> {
 
 export class NumberUintType extends UintType<number> {
   assertValidValue(value: unknown): asserts value is number {
+    if (
+      value !== Infinity 
+        && (
+          !Number.isSafeInteger(value as number)
+          || value > (BigInt(2) ** (BigInt(8) * BigInt(this.byteLength)))
+        )
+    ) {
+      throw new Error("Uint value is not a number");
+    }
     if (value as number < 0) {
       throw new Error("Uint value must be gte 0");
     }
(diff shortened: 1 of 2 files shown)


src/types/basic/uint.ts
```diff
@@ -18,9 +18,6 @@ export class UintType<T> extends BasicType<T> {
 
 export class NumberUintType extends UintType<number> {
   assertValidValue(value: unknown): asserts value is number {
-    if (!(Number.isSafeInteger(value as number) || value === Infinity)) {
```
wemeetagain, line 21: We still need this check in assertValidValue, we still only allow safe integers and infinity

src/types/basic/uint.ts
```diff
@@ -58,7 +55,10 @@ export class NumberUintType extends UintType<number> {
     return Number(output);
   }
   fromJson(data: Json): number {
-    const n = Number(data);
+    let n = Number(data);
+    if(n > Number.MAX_SAFE_INTEGER) {
+      n = Infinity;
```
wemeetagain, line 60: I think we need to be more restrictive here.
```typescript
const bigN = BigInt(data);
if (bigN === (BigInt(2) ** BigInt(this.byteLength)) - BigInt(1)) {
  n = Infinity;
} else {
  throw new Error("Uint value unsafe");
}
```
something like that, only set to Infinity if value is 0xfffffffff....

src/types/basic/uint.ts
wemeetagain: And actually, we're missing a byte length check here. eg: that `value < (BigInt(2) ** this.byteLength)`

PR #27, opened 2020-06-05
Add BN uint implementation

wemeetagain:
Add BN.js alternative Uint implementation


wemeetagain:
I added this thinking that this would alleviate cases where js environments don't support BIgInt.
Unfortunately, `ssz` relies on `persistent-merkle-tree`, which uses BigInt as well.

Closing for now, will reopen if there's a reason

PR #29, opened 2020-06-07
Implement minSize and maxSize api

### LINKED ISSUE #28: Add minSize(), maxSize() api to all types
The demand comes from eth2 spec https://github.com/ethereum/eth2.0-specs/pull/1800/files

tuyennhv:
## Goal
+ resolves #28

### DIFF
--- a/src/backings/structural/abstract.ts
+++ b/src/backings/structural/abstract.ts
@@ -22,6 +22,12 @@ export class StructuralHandler<T extends object> {
   size(target: T): number {
     throw new Error("Not implemented");
   }
+  maxSize(): number {
+    throw new Error("Not implemented.");
+  }
+  minSize(): number {
+    throw new Error("Not implemented.");
+  }
   assertValidValue(target: unknown): asserts target is T {
     throw new Error("Not implemented");
   }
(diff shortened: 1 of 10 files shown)


src/backings/structural/abstract.ts
```diff
@@ -71,4 +71,10 @@ export class StructuralHandler<T extends object> {
   toJson(value: T, options?: IJsonOptions): Json {
     throw new Error("Not implemented");
   }
+  maxSize(): number {
```
wemeetagain, line 74: can you move these methods to the type objects, eg: move from `src/backings/structural` to `src/types/composite` and `src/types/basic`? These methods would be declared right next to the `size` method for each type

tuyennhv: I moved minSize()/maxSize() next to size(). Everything started from `src/type` but for composite, it delegates its implementation to structural backing, just like `size()`

src/types/composite/abstract.ts
```diff
@@ -83,6 +83,20 @@ export class CompositeType<T extends object> {
     }
   }
 
+  /**
+   * Maximal serialized byte length
+   */
+  maxSize(): number {
+    return this.structural.maxSize();
```
tuyennhv, line 90: this is the entry point where consumers should call.

tuyennhv:
Got same result like https://gist.github.com/protolambda/db75c7faa1e94f2464787a480e5d613e

Will create unit tests in lodestar to make sure it's correct.

PR #39, opened 2020-06-27
Print type name in test declaration

dapplion:
In test logs print
```
should correctly serialize ByteVectorType
```
instead of 
```
should correctly serialize [object Object]
```


tuyennhv:
lgtm

Issue #38, opened 2020-06-27
Chained error messages


wemeetagain:
We would like errors in processing sub-fields of an ssz object to 'bubble up' to provide an error message that locates the error to the sub-field in question.

Eg:
```typescript
interface BeaconBlock {
  body: {
    attestations: {
      data: {
        slot: number;
      }
    }[]
  }
}
```
If there was an error in when running assertValidValue/deserialize of a slot inside an attestation data inside a beacon block, as in the object above, the error message would look something like: `body: attestations: index ${i}: data: slot: ${error message}`

See https://github.com/ChainSafe/ssz/blob/bbdae82194c4b0cf2a8099b99b230ff2ab2551f8/src/core/assertValidValue.ts#L120
as an example of how this was accomplished in our old ssz implementation.
Resolving this issue likely involves adding try/catch to our array and container structural/tree backing classes for a few key methods. Likely assertValidValue and deserialize/fromBytes are the most important.

wemeetagain:
cc @dapplion

PR #40, opened 2020-06-27
Track error JSON path location with try / catch

dapplion:
Attaches the location original JSON path location of an error to the error message. Fixes https://github.com/ChainSafe/ssz/issues/38
```
Error: v[1].a: Not all variable bytes consumed
      at ContainerStructuralHandler.fromBytes (src/backings/structural/container.ts:138:15)
      at ContainerStructuralHandler.deserialize (src/backings/structural/abstract.ts:62:17)
      at ContainerType.deserialize (src/types/composite/abstract.ts:110:28)
      at Context.<anonymous> (test/unit/errors.test.ts:58:27)
```

Uses a custom error SSZNestedError to keep track of the path as the Error is catched and rethrown on each nested container / array.

Let me know if there are other places where I should add the try / catch besides `ContainerStructuralHandler` and `CompositeArrayStructuralHandler`

### DIFF
--- a/src/backings/structural/array.ts
+++ b/src/backings/structural/array.ts
@@ -2,7 +2,7 @@
 import {ArrayLike, Json} from "../../interface";
 import {BasicArrayType, CompositeArrayType, IJsonOptions} from "../../types";
 import {StructuralHandler} from "./abstract";
-import {SSZNestedError} from "../../util/error";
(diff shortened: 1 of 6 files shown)


wemeetagain:
file conflict

dapplion:
> file conflict

Fixed

dapplion:
> file conflict

Fixed

PR #52, opened 2020-07-13
Container keys name checking

mpetrunic:
- when we define generator for some interface it will check if all keys are present and if there was misspell

I was hoping to fix types so it ensures you cannot assign incompatible ssz generator, but because we have any in bunch of places it always allows just BasicTypes and preventing CompositeTypes


src/types/basic/abstract.ts
```diff
@@ -165,7 +166,7 @@ export class BasicType<T> {
   /**
    * Convert to JSON-serializable object
    */
-  toJson(value: T): Json {
+  toJson(value: T, options?: IJsonOptions): Json {
```
wemeetagain, line 169: should fromJson also have these options?

mpetrunic: Probably, but none of the basic types are using it. This was just to fix some type incompatibility

wemeetagain: Can you add it for completeness if you're adding it to `toJson` here/

I'd love to refactor to merge BasicType and CompositeType at some point.

mpetrunic: Yeah we could have a lot more typesafety if we manage to merge them

mpetrunic:
No idea what's going in this PR anymore

PR #54, opened 2020-07-26
read-only tree iteration

wemeetagain:
Depends on chainsafe/persistent-merkle-tree#13 and subsequent release

- Update persistent-merkle-tree usage (`toGindex`, `iterateAtDepth` param order change)
- Use `Tree#iterateNodesAtDepth` in tree-backing `toBytes`
- Add `readOnlyForEach` and `readOnlyMap` methods and helper functions

### DIFF
--- a/src/backings/readOnlyIterate.ts
+++ b/src/backings/readOnlyIterate.ts
@@ -1,0 +1,25 @@
+import {ArrayLike} from "../interface";
+
+export function readOnlyForEach<T>(value: ArrayLike<T>, fn: (value: T, index: number) => void): void {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  if ((value as any).readOnlyForEach) {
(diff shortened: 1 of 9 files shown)


src/backings/readOnlyIterate.ts
```diff
@@ -0,0 +1,25 @@
+import {ArrayLike} from "../interface";
+
+export function readOnlyForEach<T>(value: ArrayLike<T>, fn: (value: T, index: number) => void): void {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  if ((value as any).readOnlyForEach) {
```
mpetrunic, line 5: Can you create ReadOnlyIterator<T> type?
```typescript
ReadOnlyIterator<T> {
    readOnlyForEach: (value: T, index: number) => void;
    readOnlyMap<U>: (value: T, index: number) => U:
}

```


You would have `value: ArrayLike<T>|ReadOnlyIterator<T>` without any casting

PR #56, opened 2020-07-31
Optimize fromStructural, use subtreeFillToContents

wemeetagain:
~Depends on https://github.com/ChainSafe/persistent-merkle-tree/pull/15 and subsequent release~

`fromStructural` is the method to convert structural objects to tree objects

Previously, `fromStructural` has the following strategy:
- create a default tree, loop thru the structural values, setting each sub-value to each place in the tree

Now:
- loop thru the structural values, creating an array of leaf nodes, create a tree from the leaf nodes

This new approach is significantly faster, as it avoids navigation thru the tree + rebinding nodes for each sub-value.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "types": "lib/index.d.ts",
   "dependencies": {
     "@chainsafe/as-sha256": "^0.2.0",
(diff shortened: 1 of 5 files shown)


wemeetagain:
ready for review

PR #64, opened 2020-12-11
Add methods to obtain generalized index of a type

mpetrunic:
As per specified here: https://github.com/ethereum/eth2.0-specs/blob/vbuterin-patch-2/ssz/merkle-proofs.md#ssz-object-to-index

I've changed some existing implementation to align with spec, so I would ask you to pay close attention.
I'm happy to jump on a call if something is unclear or just to go trough the code.

There is a lot of duplicated code since list/vector doesn't have common ancenstor. We could try to address that in separate PR.
Happy to take ideas what's the best way to test this without manually calculating index for every test case 😄 

TODO:
- [ ] tests

### DIFF
--- a/src/types/composite/container.ts
+++ b/src/types/composite/container.ts
@@ -59,6 +59,6 @@ export class ContainerType<T extends ObjectLike = ObjectLike> extends CompositeT
     const [pos] = this.getItemPosition(path);
     const baseIndex = 1;
     root = root * baseIndex * getPowerOfTwoCeil(this.chunkCount()) + pos;
-    return this.fields[path].getGeneralizedIndex(root, ...path.slice(1));
+    return this.fields[path].getGeneralizedIndex(root, ...paths.slice(1));
   }
 }
--- a/src/types/composite/container.ts
+++ b/src/types/composite/container.ts
@@ -45,11 +45,11 @@ export class ContainerType<T extends ObjectLike = ObjectLike> extends CompositeT
     return [pos, 0, this.fields[propertyName].getItemLength()];
   }
 
-  getGeneralizedIndex(root = 1, ...paths: GIndexPathKeys[]): number {
-    if (paths.length === 0) {
-      return root;
+  getGeneralizedIndex(pathParts: GIndexPathKeys[], rootIndex = 1): number {
+    if (pathParts.length === 0) {
+      return rootIndex;
     }
-    const path = paths[0];
+    const path = pathParts[0];
     if (typeof path !== "string" || path === GINDEX_LEN_PATH) {
       throw new Error(`Unsupported path ${path} in container`);
     }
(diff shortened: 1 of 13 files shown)


mpetrunic:
I guess some unit tests are failing because I changed some chunk_counts. Not sure if spec changes for lightclient  or just some diversion from spec in rest of code.

dapplion:
There weren't any tests that covers cases showing that it didn't follow the spec?

mpetrunic:
I was hoping for @wemeetagain 's feedback. I would bet it has something to do with how we handle bits in javascript. So maybe my change was wrong because I followed spec code which isn't same as ours.

src/types/composite/container.ts
```diff
@@ -1,36 +1,64 @@
...
   return isTypeOf(type, CONTAINER_TYPE);
 }
-
 export class ContainerType<T extends ObjectLike = ObjectLike> extends CompositeType<T> {
   // ES6 ensures key order is chronological
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  fields: Record<string, Type<any>>;
-  constructor(options: IContainerOptions) {
+  fields: Record<keyof T, Type<any>>;
+  constructor(options: IContainerOptions<T>) {
```
wemeetagain, line 23: Can you split out these type updates to a separate PR?
This can be merged pretty quickly

wemeetagain (commented):
After looking at this PR, it is clear to me that our impl needs a bit of cleanup.
Specifically, some of the methods that live in `src/backings/tree` (regarding gindex calculations, etc) should be moved to the type itself. (our persistent-merkle-tree library could also use a few of the gindex helper methods, like `concat_gindex`).

We currently have almost all the functionality (other than the top-level `getGeneralizedIndex` method) in the tree-backing code.  IMO we should update the library to better use what we already have. (I can help with that)

Generally, our implementation does things a little different than the spec. Think naive state transition vs fast state transition. For example, the offsets returned from the `getItemPosition` function from the spec aren't likely useful to us, because we have segregated our types so that we only need to handle chunk offsets in basic arrays, the other types always deal in whole chunks. Its not that chunk offsets aren't important, its that a function with the interface of `getItemPosition` likely isn't actually useful -- the spec contains a _naive_ implementation. IMO we don't actually need the function lifted directly from the spec. Likewise, our implementation of `getGeneralizedIndex` won't/shouldn't need to use `getItemPosition`, like the spec does.

src/types/composite/container.ts
mpetrunic: Sure, I'm gonna try to introduce abstract and some more codesharing between basic and composite array and test against lodestar

mpetrunic:
I've created python script to generate test cases for generalized index: https://github.com/ChainSafe/eth2-gindex-tests

Gonna try to modify our implementation to follow proto's remerkeable implementation and saatisfy added tests

dapplion:
The ssz implementation has changed significantly since this PR. Closing for clean-up purposes of old PRs. Happy to revisit this in the future

PR #66, opened 2020-12-16
Better abstractions and types

mpetrunic:
Still wip but provides basic frame to start sharing more code and have better typesafety when dealing with ssz.

TODO:
- figure out error when running tests


src/backings/structural/container.ts
```diff
@@ -74,18 +77,20 @@ export class ContainerStructuralHandler<T extends ObjectLike> extends Structural
...
   clone(value: T): T {
     const newValue = {} as T;
     Object.entries(this._type.fields).forEach(([fieldName, fieldType]) => {
       if (fieldType.isBasic()) {
-        newValue[fieldName as keyof T] = fieldType.clone(value[fieldName]);
-      } else {
-        newValue[fieldName as keyof T] = fieldType.structural.clone(value[fieldName]);
+        newValue[fieldName as keyof T] = fieldType.clone(value[fieldName]) as T[keyof T];
+      }
+      if (fieldType.isComposite()) {
```
wemeetagain, line 92: why is this needed now?

src/backings/structural/container.ts
```diff
@@ -128,20 +133,21 @@ export class ContainerStructuralHandler<T extends ObjectLike> extends Structural
           if (offsets[offsetIndex] > offsets[offsetIndex + 1]) {
             throw new Error("Offsets must be increasing");
           }
-          value[fieldName as keyof T] = (fieldType as CompositeType<T[keyof T]>).structural.fromBytes(
+          value[fieldName as keyof T] = (fieldType as CompositeType<T>).structural.fromBytes(
```
wemeetagain, line 136: `CompositeType<object>` or something, its definitely not `CompositeType<T>`

src/util/types.ts
```diff
@@ -0,0 +1 @@
+export type ArrayElement<T> = T extends ArrayLike<infer U> ? U : T;
```
wemeetagain, line 1: Is this used anywhere?

src/types/basic/abstract.ts
```diff
@@ -20,36 +19,15 @@ export function isTypeOf(type: unknown, typeSymbol: symbol): boolean {
...
-export class BasicType<T> {
-  /**
-   * Symbols used to track the identity of a type
-   *
-   * Used by various isFooType functions
-   */
-  _typeSymbols: Set<symbol>;
-
+export abstract class BasicType<T> extends Type<T> {
   constructor() {
```
wemeetagain, line 23: Can remove constructor

src/types/composite/abstract.ts
```diff
@@ -1,29 +1,18 @@
...
 import {BasicType} from "../basic";
-import {IJsonOptions} from "../type";
+import {IJsonOptions, Type} from "../type";
 
 /**
  * A CompositeType is a type containing other types, and is flexible in its representation.
  *
  */
-export class CompositeType<T extends object> {
+export abstract class CompositeType<T extends CompositeValue = ObjectLike> extends Type<T> {
```
wemeetagain, line 11: no default type here?

src/types/composite/array.ts
```diff
@@ -8,18 +8,18 @@ export interface IArrayOptions {
...
   constructor(options: IArrayOptions) {
     super();
     this.elementType = options.elementType as BasicType<T>;
   }
 }
 
-export class CompositeArrayType<T extends ArrayLike<unknown>> extends CompositeType<T> {
-  elementType: CompositeType<object>;
+export abstract class CompositeArrayType<T extends ArrayLike<unknown>> extends CompositeType<T> {
+  elementType: CompositeType<Record<string, unknown>>;
```
wemeetagain, line 20: CompositeValue

src/types/composite/array.ts
```diff
@@ -8,18 +8,18 @@ export interface IArrayOptions {
...
   constructor(options: IArrayOptions) {
     super();
-    this.elementType = (options.elementType as unknown) as CompositeType<object>;
+    this.elementType = (options.elementType as unknown) as CompositeType<Record<string, unknown>>;
```
wemeetagain, line 23: CompositeValue

src/types/composite/container.ts
```diff
@@ -4,9 +4,8 @@ import {isTypeOf} from "../basic";
 import {Type} from "../type";
 import {ContainerStructuralHandler, ContainerTreeHandler, ContainerByteArrayHandler} from "../../backings";
 
-export interface IContainerOptions {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  fields: Record<string, Type<any>>;
+export interface IContainerOptions<T extends Record<string, unknown>> {
```
wemeetagain, line 7: extends ObjectLike

src/types/composite/container.ts
```diff
@@ -28,7 +26,7 @@ export class ContainerType<T extends ObjectLike = ObjectLike> extends CompositeT
     this._typeSymbols.add(CONTAINER_TYPE);
   }
   isVariableSize(): boolean {
-    return Object.values(this.fields).some((fieldType) => fieldType.isVariableSize());
+    return Object.values(this.fields as Record<string, Type<unknown>>).some((fieldType) => fieldType.isVariableSize());
```
wemeetagain, line 29: Is this type coersion needed?

src/types/composite/root.ts
```diff
@@ -1,21 +1,22 @@
 import {isTypeOf} from "../basic";
 import {ByteVectorType} from "./byteVector";
 import {CompositeType} from "./abstract";
+import {ObjectLike} from "../../interface";
 
 /**
  * Allow for lazily evaulated expandedType thunk
  */
-export interface IRootOptions<T extends object> {
+export interface IRootOptions<T extends ObjectLike> {
```
wemeetagain, line 9: In this file, CompositeValue

src/types/composite/vector.ts
```diff
@@ -57,7 +57,7 @@ export class BasicVectorType<T extends Vector<unknown> = Vector<unknown>> extend
   }
 }
 
-export class CompositeVectorType<T extends Vector<object> = Vector<object>> extends CompositeArrayType<T> {
+export class CompositeVectorType<T extends Vector<ObjectLike> = Vector<ObjectLike>> extends CompositeArrayType<T> {
```
wemeetagain, line 60: In this file, CompositeValue

src/types/type.ts
```diff
@@ -1,11 +1,78 @@
...
-export type Type<T> = BasicType<T> | (T extends object ? CompositeType<T> : never);
-
 export interface IJsonOptions {
   case: "camel" | "snake";
 }
+
+// /**
+//  * A Type is either a BasicType or a CompositeType.
+//  */
+// export type Type<T> = BasicType<T> | (T extends object ? CompositeType<T> : never);
```
wemeetagain, line 12: I think we could move this code into the jsdoc

src/backings/structural/vector.ts
```diff
@@ -54,7 +54,7 @@ export class BasicVectorStructuralHandler<T extends Vector<unknown>> extends Bas
   }
 }
 
-export class CompositeVectorStructuralHandler<T extends Vector<object>> extends CompositeArrayStructuralHandler<T> {
+export class CompositeVectorStructuralHandler<T extends Vector<ObjectLike>> extends CompositeArrayStructuralHandler<T> {
```
wemeetagain, line 57: CompositeValue

src/backings/structural/list.ts
```diff
@@ -49,7 +49,7 @@ export class BasicListStructuralHandler<T extends List<unknown>> extends BasicAr
   }
 }
 
-export class CompositeListStructuralHandler<T extends List<object>> extends CompositeArrayStructuralHandler<T> {
+export class CompositeListStructuralHandler<T extends List<ObjectLike>> extends CompositeArrayStructuralHandler<T> {
```
wemeetagain, line 52: CompositeValue

src/backings/structural/container.ts
```diff
@@ -1,10 +1,10 @@
...
-import {StructuralHandler} from "./abstract";
-import {toExpectedCase} from "../utils";
+import {Json, ObjectLike} from "../../interface";
+import {CompositeType, ContainerType, IJsonOptions} from "../../types";
 import {SszErrorPath} from "../../util/errorPath";
+import {toExpectedCase} from "../utils";
+import {StructuralHandler} from "./abstract";
 
-export class ContainerStructuralHandler<T extends ObjectLike> extends StructuralHandler<T> {
+export class ContainerStructuralHandler<T extends Record<string, unknown>> extends StructuralHandler<T> {
```
wemeetagain, line 7: ObjectLike

src/backings/byteArray/list.ts
```diff
@@ -17,7 +17,7 @@ export class BasicListByteArrayHandler<T extends List<unknown>> extends BasicArr
   }
 }
 
-export class CompositeListByteArrayHandler<T extends List<object>> extends CompositeArrayByteArrayHandler<T> {
+export class CompositeListByteArrayHandler<T extends List<ObjectLike>> extends CompositeArrayByteArrayHandler<T> {
```
wemeetagain, line 20: CompositeValue

src/backings/byteArray/vector.ts
```diff
@@ -13,7 +13,7 @@ export class BasicVectorByteArrayHandler<T extends Vector<unknown>> extends Basi
   }
 }
 
-export class CompositeVectorByteArrayHandler<T extends Vector<object>> extends CompositeArrayByteArrayHandler<T> {
+export class CompositeVectorByteArrayHandler<T extends Vector<ObjectLike>> extends CompositeArrayByteArrayHandler<T> {
```
wemeetagain, line 16: CompositeValue

wemeetagain (changes requested):
Looks really good :)
Good cleanup types with `CompositeValue` and `ObjectLike` type and `Type` abstract class.

Most comments are minor type changes. (all instances of `object` should be changed to `CompositeValue`) (`Record<string, unknown>` should be replaced with `ObjectLike` where possible)

Only issue I have is re: usage of `isComposite`.
I think its a useful method for consumers, but internal to the library, IMO its not necessary and should be replaced with type coersion as necessary.
SSZ has several invariants that don't need to be checked at runtime (basic types are always fixed-length, types are always either basic or composite)

src/backings/structural/container.ts
mpetrunic: Because typescript is stupid. It cannot (or at least I didn't find a way) to infer is isBasic is false that type is composite. 

That said when going trough code. It was a lot easier to understand code with usage of isComposite to I decided to leave it on. We can make it protected so end users don't see it.

src/util/types.ts
mpetrunic: I think it is. I was trying to bring even more type safety to infer element type from List and Vector. Will check

src/types/basic/abstract.ts
mpetrunic: Yup. I've put it because of wierd constructor error. But it was unrelated

src/types/composite/container.ts
mpetrunic: need to check. I think I can't use keyof with ObjectLike

src/types/composite/container.ts
mpetrunic: unfortunately yes. Because interfaces don't have index signature, we have to add `{}` which casuses fields to be Type | never . Never is the case if there is no key in given object. I have no idea how to fix that like we cannot create generic that accepts every interface unless we relax it with `{}` which basically allows anything

src/backings/structural/container.ts
```diff
@@ -58,10 +59,12 @@ export class ContainerStructuralHandler<T extends ObjectLike> extends Structural
     Object.entries(this._type.fields).forEach(([fieldName, fieldType]) => {
       try {
         if (fieldType.isBasic()) {
-          (fieldType as Type<T[keyof T]>).assertValidValue((value as T)[fieldName]);
-        } else {
           // @ts-ignore
-          fieldType.structural.assertValidValue((value as T)[fieldName]);
+          fieldType.assertValidValue(value[fieldName]);
+        }
+        if (fieldType.isComposite()) {
```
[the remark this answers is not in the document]
mpetrunic: If you wanna add another like in this code block, it will be typesafe. I have no idea how to fix wierd error with that assertValidValue. Tried every combination. I guess you had same one since it was ignored before 😄

src/backings/structural/array.ts
```diff
@@ -88,7 +88,7 @@ export class BasicArrayStructuralHandler<T extends ArrayLike<unknown>> extends S
   }
 }
 
-export class CompositeArrayStructuralHandler<T extends ArrayLike<object>> extends StructuralHandler<T> {
+export class CompositeArrayStructuralHandler<T extends ArrayLike<ObjectLike>> extends StructuralHandler<T> {
```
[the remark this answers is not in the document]
mpetrunic: If I put CompositeValue, that allows array element to be another array, which is perfectly fine. But our types doesn't reflect that anywhere so there is huge number of type errors. It's kinda hard because wherever I change type I cause chain reaction.

Not sure how to proceed. Maybe we can in separate PRs fix types from bottom up?

src/backings/structural/container.ts
mpetrunic: that screws everything since it ObjectLike allows empty object without keys which casues keyof to stop working

dapplion:
Good ideas in this PR! But I would close for now since the codebase has changed substantially since https://github.com/ChainSafe/ssz/pull/82. If this effort is gonna be pursued in the future better start from a new branch from current master

PR #72, opened 2021-02-09
Restrict the use of BigInt whenever possible

### LINKED ISSUE #71: Avoid using BigInt whenever possible
v8 has to allocate an object for each BigInt under the cover: https://v8.dev/blog/bigint and it causes memory/performance issue in lodestar, so we need to use Number instead whenever possible.
BigInt-memory-issue

also we need to add a performance test for `uint.ts` `toBytes(), fromBytes()`
```ts
  it.only("reproduce readOnlyMap issue", async function () {
    this.timeout(0);
    let minTime = Number.MAX_SAFE_INTEGER;
    let maxTime = 0;
    const originalState = state.getOriginalState();
    for (let i = 0; i < 20000; i++) {
      const start = Date.now();
      readOnlyMap(originalState.balances, (balance) => balance);
      const duration = Date.now() - start;
      if (duration < minTime) minTime = duration;
      if (duration > maxTime) maxTime = duration;
    }
    console.log("@@@ minTime", minTime, "maxTime", maxTime);
  });
```

this currently prints out
```
2021-02-09 08:15:42 []                 info: Loaded state slot=756416, numValidators=114038
@@@ minTime 140 maxTime 1109
```

tuyennhv:
resolves #71

+ process Uint types by a group of 4 bytes (which supports shift operators for number), so we only use BigInt once per 4 bytes (25% compared to the current implementation)
+ Added performance tests

|test|minTime|averageTime|maxTime|
|----|--------|-------------|---------|
|master - serialize|228|262|745|
|this branch - serialize|90|100|480|
|master - deserialize|241|278|663|
|this branch - deserialize|85|100|320|
|master - readOnlyMap perf test|141|173|636|
|this branch - readOnlyMap perf test|39|55|262|

+ passed `test:ssz` and `test:ssz-min` spec tests in lodestar

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "prepublishOnly": "yarn build",
     "pretest": "yarn check-types",
     "test": "yarn test:unit",
+    "test:perf": "mocha --colors -r ts-node/register \"test/perf/**/*.test.ts\"",
     "test:unit": "nyc -e .ts mocha --colors -r ts-node/register \"test/unit/**/*.test.ts\""
   },
   "types": "lib/index.d.ts",
--- a/src/types/basic/uint.ts
+++ b/src/types/basic/uint.ts
@@ -30,6 +30,9 @@ export function isNumberUintType(type: unknown): type is NumberUintType {
   return isTypeOf(type, NUMBER_UINT_TYPE);
 }
 
+// a group contains 4 bytes, that's what << operator supports for number
+const BYTES_PER_GROUP = 4;
+
 export class NumberUintType extends UintType<number> {
   constructor(options: IUintOptions) {
     super(options);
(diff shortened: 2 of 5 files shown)


tuyennhv:
added performance test statistics, ready for review

wemeetagain:
You should try using `Uint32Array`, `BigUint64Array`, or `DataView` to do the grouping for you. Might need to be `DataView` to have control over the endianness.

tuyennhv:
> You should try using `Uint32Array`, `BigUint64Array`, or `DataView` to do the grouping for you. Might need to be `DataView` to have control over the endianness.

thanks for the suggestion, I tried `Uint32Array` and `DataView`, it really makes the code shorter and more straightforward, however there's a small concern about the performance:
+ in average, using `Uint32Array` is 10% slower than managing byte grouping ourself, and using `DataView` is 10% slower than `Uint32Array`
+ the average max time in performance test of `Uint32Array` and `DataView` (~470ms) is 50% slower than managing byte grouping ourself (~320ms). On master, max time is ~660ms.

since we focus on performance to help mitigating event loop lag issue in lodestar, I'm leaning toward managing byte grouping ourself. Please let me know your ideas.

wemeetagain (approved):

wemeetagain (approved):

PR #74, opened 2021-02-10
Use abstract class Type/BasicType/CompositeType

wemeetagain:
BREAKING CHANGE:
Remove Type#isBasic in favor of isBasicType function

- Make `Type` an abstract class (rather than a type)
- Make `BasicType`, `CompositeType`, , `UintType`, `BasicArrayType`, `CompositeArrayType` abstract
- `Type#isBasic` now no longer works as intended (to coerce the type to a `CompositeType`), so remove it, use `isBasicType`/`isCompositeType` instead

Any consumers that previously used `Type#isBasic` can now use either `isBasicType` or `isCompositeType`
```ts

const t: Type<unknown> = ...;

// before

if (t.isBasic()) {
  ...
} else {
  // t is a CompositeType
  t.tree.defaultValue();
}

// after

if (!isCompositeType(t)) {
  ...
} else {
  // t is a CompositeType
  t.tree.defaultValue();
}
```

### DIFF
--- a/src/backings/structural/container.ts
+++ b/src/backings/structural/container.ts
@@ -1,5 +1,5 @@
 import {ObjectLike, Json} from "../../interface";
-import {ContainerType, CompositeType, Type, IJsonOptions} from "../../types";
+import {ContainerType, CompositeType, IJsonOptions, isCompositeType} from "../../types";
 import {StructuralHandler} from "./abstract";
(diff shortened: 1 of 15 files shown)


mpetrunic (approved):

mpetrunic:
Seems like ssz repo is missing eslint rules for unsused imports ^^

PR #79, opened 2021-02-17
Improve BigIntUintType.toBytes() to use number bitshifting

### LINKED ISSUE #77: Improve `balances` setter for BeaconState
part of https://github.com/ChainSafe/lodestar/issues/2046

this is a quick performance test I have

```ts
import { BeaconState } from "./objects";

describe("fromStructural", function () {
  it.only("reproduce the event loop lag", function () {
    this.timeout(0);
    const balances = Array.from({length: 200000}, () => BigInt(31217089836));
    let minTime = Number.MAX_SAFE_INTEGER;
    let maxTime = 0;
    let average = {duration: 0, count: 0};
    const MAX_TRY = 10000;
    for (let i = 0; i < MAX_TRY; i++) {
      const state = BeaconState.tree.defaultValue();
      const start = Date.now();
      state.balances = balances;
      const duration = Date.now() - start;
      const totalDuration = average.duration * average.count + duration;
      const totalCount = average.count + 1;
      average.count = totalCount;
      average.duration = totalDuration / totalCount;
      if (duration < minTime) minTime = duration;
      if (duration > maxTime) maxTime = duration;
    }
    console.log("fromStructural minTime:", minTime, "maxTime:", maxTime, "average:", average.duration, "MAX_TRY:", MAX_TRY);
  });
});
```

this prints out 
+ 1st time `fromStructural minTime: 262 maxTime: 999 average: 307.5632000000007 MAX_TRY: 10000`
+ 2nd time `fromStructural minTime: 258 maxTime: 676 average: 291.47639999999956 MAX_TRY: 10000`

tuyennhv:
resolves #77 

+ Switch back to use number bit shifting to improve performance. Added `fromStructural.test.ts` to prove that and for other improvements in the future if any
+ Leverage `BigInt.asUintN` api and added `BIGINT_4_BYTES` constant to reduce BigInt allocation

Performance statistics for `fromStructural` test:

|Branch|minTime (ms)|maxTime (ms)|average (ms)|
|-------|--------------|--------------|-------------|
|master - 1st|105|678|124.9|
|master - 2nd|104|583|119.32|
|this PR - 1st|62|322|78.00|
|this PR - 2nd|64|376|78.61|

### DIFF
--- a/src/types/basic/uint.ts
+++ b/src/types/basic/uint.ts
@@ -139,9 +139,10 @@ export class BigIntUintType extends UintType<bigint> {
     let v = value;
     let groupedBytes = Number(BigInt.asUintN(32, v));
     for (let i = 0; i < this.byteLength; i++) {
-      const byteIndex = i % 4;
-      output[offset + i] = Number((groupedBytes >> (8 * byteIndex)) & 0xff);
-      if ((i + 1) % 4 === 0) {
+      output[offset + i] = Number(groupedBytes & 0xff);
+      if ((i + 1) % 4 !== 0) {
+        groupedBytes >>= 8;
+      } else {
         v >>= BIGINT_4_BYTES;
         groupedBytes = Number(BigInt.asUintN(32, v));
       }
--- a/src/types/basic/uint.ts
+++ b/src/types/basic/uint.ts
@@ -27,6 +27,8 @@ export abstract class UintType<T> extends BasicType<T> {
 
 export const NUMBER_UINT_TYPE = Symbol.for("ssz/NumberUintType");
 
+const BIGINT_4_BYTES = BigInt(32);
+
 export function isNumberUintType(type: Type<unknown>): type is NumberUintType {
   return isTypeOf(type, NUMBER_UINT_TYPE);
 }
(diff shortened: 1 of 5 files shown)


dapplion:
For benchmarking time I would recommend you use this API which is much more precise than Date.now() https://nodejs.org/api/process.html#process_process_hrtime_bigint

wemeetagain:
Wowza, this is awesome :rocket:

tuyennhv:
> https://nodejs.org/api/process.html#process_process_hrtime_bigint

so I tried the api, while `minTime` and `averageTime` are the same, the `maxTime` tends to be bigger: 474ms, 417ms and 343ms. I guess it's because we do BigInt operations using that api while BigInt allocation was already a concern to us.

I think we should keep `Date.now()` with `ms` for now. For any performance tests with <5ms, we can think about applying `process.hrtime.bigint()`.

dapplion:
@tuyennhv To benchmark what you could do is
```ts
const ops = 100;
const data = prepareData(ops)
const from = process.hrtime.bigint()
for (let i=0; i<ops; i++) {
  fnToBenchmark(data[i])
}
const to = process.hrtime.bigint()
const timePerOp = (to - from) / ops
```
Then you even out the overhead around timing

tuyennhv:
> @tuyennhv To benchmark what you could do is
> 
> ```ts
> const ops = 100;
> const data = prepareData(ops)
> const from = process.hrtime.bigint()
> for (let i=0; i<ops; i++) {
>   fnToBenchmark(data[i])
> }
> const to = process.hrtime.bigint()
> const timePerOp = (to - from) / ops
> ```
> 
> Then you even out the overhead around timing

this is good for `averageTime` but not for `minTime` and `maxTime`

dapplion:
> this is good for `averageTime` but not for `minTime` and `maxTime`

Right, but due to the unpredictable nature of garbage collection in javascript, is `maxTime` a number we can trust indicative or problems?

tuyennhv:
> > this is good for `averageTime` but not for `minTime` and `maxTime`
> 
> Right, but due to the unpredictable nature of garbage collection in javascript, is `maxTime` a number we can trust to be indicative or problems?

+ `maxTime` is needed to know the worse case, it's usually close to the long epoch transition we saw in the profiler so I want to see it in all of the tests. It makes me feel confident that we can reproduce the long epoch transition in the performance test.
+ although in this scenario it's not predictable due to gc, this PR clearly shows that maxTime is improved compared to master so I still can trust it (although I usually runs the test multiple times to make sure). We need it for future improvement too.
+ the performance in general is not always unpredictable due to gc. In the profiler `getAttestationDeltas` runs very slowly at that time but there is no `gc` at that time

tuyennhv:
if I use wasm, it's improved a little bit with average performance like `minTime 42, maxTime 436, average 56`. Note that I have to use nodejs `--experimental-wasm-bigint` to pass bigint to wasm.

wemeetagain (approved):

PR #81, opened 2021-03-04
Use bitshifting whenever possible

tuyennhv:


tuyennhv:
it turns out bit shifting only support 32 bytes

PR #82, opened 2021-03-08
Refactor codebase

wemeetagain:
- Move all struct and tree logic to the `Type`
  - Instead of having separate 'handler' instances with circular references back to the type, attach all methods to the type
- Rename all underlying struct and tree logic with `struct_*` and `tree_*`
  - High-level wrapper functions and backing-agnostic methods remain un-prefixed (eg: `getMaxSerializedLength`)
  - `struct_*` methods operate on structs, using type information + get/set to perform ssz operations
  - `tree_*` methods operate on `Tree`s, using type information to perform ssz operations
  - `tree_getProperty` and `tree_setProperty` are high level methods that may either return a primitive or a `Tree`, depending on the property type
    - This is the primary distinction between `BasicArrayType` (which always gets/sets primitives), `CompositeArrayType` (which always gets/sets trees), and `ContainerType` (which gets/sets primitives or trees depending on the property type)
- Simplify tree-backed proxy logic
  - Instead of the tree logic instance doubling as a proxy handler, separate out proxy logic as a single object
  - `TreeValue` and its subclasses create a convenience wrapper around low level `Type.tree_*` functions, are wrapped in a proxy for convenient get/set
  - use `createTreeBacked(type: CompositeType<T>, tree: Tree): TreeBacked<T>` function to create `TreeBacked` objects

Benefits/tradeoffs of this refactor:
- Circular references in composite types were confusing, much of the previous logic looked like `this._type.structural.foo()`
  - Code split into structural / tree / byte array "handlers" is now unified on the Type. This leads to simpler code, but more code per file.
- Proxy logic now in a single place w/ simpler proxy logic. Previous implementation combined proxy handler with tree logic.
  - ITreeBacked methods were previously implemented as curried functions, always bound with a tree. This was very confusing and sometimes led to methods defined with a tree parameter that was never used.
- very few breaking changes (anywhere we used `type.tree` previously)

Example code:
```ts
import {phase0} from "@chainsafe/lodestar-types";
import {createTreeBacked} from "@chainsafe/ssz";

// tree_* methods operate on Trees
const tree = phase0.BeaconState.tree_defaultValue();
phase0.BeaconState.tree_serialize(tree);
const genesisTime = phase0.BeaconState.tree_getProperty(tree, "genesisTime"); // returns a number
const validators = phase0.BeaconState.tree_getProperty(tree, "validators"); // returns a Tree
phase0.BeaconState.tree_setProperty(tree, "genesisTime", 5);

// but usually you'd create a tree-backed beacon state
const state = phase0.BeaconState.createTreeBacked(tree);

state.genesisTime; // -> number
state.validators; // -> TreeBacked<List<Validator>>
state.genesisTime = 5;

// get the backing tree
state.tree
// get the backing type
state.type
// any ITreeBacked method
state.hashTreeRoot();
state.createProof([["validators", 0, "effectiveBalance"]]);
state.serialize();
state.entries();

// struct_* methods operate on structs
const struct = phase0.BeaconState.struct_defaultValue();
phase0.BeaconState.struct_hashTreeRoot(struct);
// but most of the struct_* methods are aliased without this prefix (and will also try to do the 'smart' thing with tree-backed values)
phase0.BeaconState.defaultValue();
phase0.BeaconState.hashTreeRoot(struct);
```


src/backings/tree.ts
```diff
@@ -0,0 +1,376 @@
...
+import {isTree} from "../util/tree";
+import {byteArrayEquals} from "../util/byteArray";
+import {Path} from "./backedValue";
+
+export function isTreeBacked<T extends CompositeValue>(value: unknown): value is ITreeBacked<T> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return value && (value as any).type && (value as any).tree && isTree((value as any).tree);
+}
+
+export type KeyOf = string | number;
```
mpetrunic, line 25: Since it's exported might be beteer in separate file

src/backings/tree.ts
```diff
@@ -0,0 +1,376 @@
...
+export type ValueOf<T extends CompositeValue, V extends keyof T = keyof T> = ITreeBacked<T[V]> | T[V];
```
mpetrunic, line 26: same as above

src/backings/tree.ts
```diff
@@ -0,0 +1,376 @@
...
+   * The merkle tree backing
+   */
+  tree: Tree;
+
+  /**
+   * Equality
+   *
+   * If both values are tree-backed, use equality by merkle root, else structural equality
+   */
+  equals(other: T): boolean;
```
mpetrunic, line 47: shouldn't other be of type ITreeBacked or unknown?

src/backings/tree.ts
```diff
@@ -0,0 +1,376 @@
...
+  serialize(): Uint8Array;
+
+  /**
+   * Merkleization
+   */
+  hashTreeRoot(): Uint8Array;
+
+  createProof(paths: Path[]): Proof;
+
+  keys(): IterableIterator<KeyOf>;
```
mpetrunic, line 75:
```suggestion
  keys(): IterableIterator<keyof T>;
```
?

src/backings/tree.ts
```diff
@@ -0,0 +1,376 @@
...
+   * Serialization
+   */
+  serialize(): Uint8Array;
+
+  /**
+   * Merkleization
+   */
+  hashTreeRoot(): Uint8Array;
+
+  createProof(paths: Path[]): Proof;
```
mpetrunic, line 73: It would be supper awesome if we could make path typesafe. But it would require some recursive keyof and wouldn't work with arrays right?

src/backings/tree.ts
```diff
@@ -0,0 +1,376 @@
...
+    const keys = this.getPropertyNames();
+    let i = 0;
+    for (const value of this.values()) {
+      yield [keys[i], value];
+      i++;
+    }
+  }
+}
+
+export class BasicListTreeValue<T extends List<unknown>> extends BasicArrayTreeValue<T> {
```
mpetrunic, line 283: it would be way more readable if each class is in separaet file inside tree directory

src/types/basic/abstract.ts
```diff
@@ -19,98 +20,52 @@ export abstract class BasicType<T> extends Type<T> {
...
+    return this.struct_getSerializedLength();
   }
 
-  /**
-   * Validate bytes before calling fromBytes
-   * @param data
-   * @param offset
-   */
-  validateBytes(data: Uint8Array, offset: number): void {
+  bytes_validate(data: Uint8Array, offset: number): void {
```
mpetrunic, line 50: why no camelcase?

src/types/composite/abstract.ts
```diff
@@ -1,34 +1,206 @@
...
 /**
  * A CompositeType is a type containing other types, and is flexible in its representation.
  *
  */
-export abstract class CompositeType<T extends object> extends Type<T> {
-  structural: StructuralHandler<T>;
-  tree: TreeHandler<T>;
-  byteArray: ByteArrayHandler<T>;
+export abstract class CompositeType<T extends CompositeValue> extends Type<T> {
+  _chunkDepth: number;
```
mpetrunic, line 31: protected?

src/types/composite/abstract.ts
```diff
@@ -1,34 +1,206 @@
...
+  _defaultNode: Node;
```
mpetrunic, line 32: protected?

src/backings/tree.ts
dapplion: Yes it does! This compiles

```ts
type State = {
  a: {
    b: {
      c: {
        d: string;
      };
    }[];
  };
};

gett({} as State, "a", "b", 1, "c");

function gett<
  T,
  K1 extends keyof T,
  K2 extends keyof T[K1],
  K3 extends keyof T[K1][K2],
  K4 extends keyof T[K1][K2][K3]
>(val: T, k1: K1, k2: K2, k3: K3, k4: K4) {}
```

src/backings/tree.ts
```diff
@@ -0,0 +1,376 @@
...
+export class CompositeListTreeValue<T extends List<object>> extends CompositeArrayTreeValue<T> {
+  type: CompositeListType<T>;
+
+  constructor(type: CompositeListType<T>, tree: Tree) {
+    super(type, tree);
+    this.type = type;
+  }
+
+  push(...values: ValueOf<T>[]): number {
+    const convertedValues = Array.from({length: values.length}, (_, i) => {
```
dapplion, line 308: Why no use `Array.prototype.map()` here?

src/backings/tree.ts
wemeetagain: I changed to separate files, now tests fail due to circular import :( plz help

src/backings/tree.ts
wemeetagain: I think type `T` makes sense here. `type TreeBacked<T> = ITreeBacked<T> & T`, so `TreeBacked` values will be included.

src/backings/tree.ts
mpetrunic: I can maybe look into that tommorow

mpetrunic:
I've used https://www.npmjs.com/package/madge to generate:
graph

Ugh, this is a bit annoying.

abstract <-> proxy is easy to resolve. Just create ITreeValue interface in interface.ts with properties required by proxy.

Circular references between proxy and implementations I'm not sure. It might work if abstract doesn't depend on proxy :/

dapplion:
Could we prioritize merging the PR and then finding a way to split tree in multiple files? Would be really nice to have this merged

dapplion (approved):

wemeetagain:
I'm going to fully integrate with lodestar and run spec tests before merging.

dapplion (approved):

wemeetagain:
Integrated with Lodestar, passing all tests - unit, e2e, sim, spec, etc.

PR #89, opened 2021-04-01
Improve convertToTree

tuyennhv:
part of https://github.com/ChainSafe/lodestar/issues/2318

**Motivation**

`PendingAttestations` has `aggregationBits` and `data` as TreeBacked values (because they stay inside `attestations` of `TreeBacked<SignedBeaconBlock>`) but it takes so much time to create tree.

**Description**
+ If a value is tree backed, `convertToTree` should return the tree immediately.
+ Reproduce the issue in newly created `convertToTree` performance test.

**Performance test result**
+ master: 368ms
+ this PR: 18ms


wemeetagain (approved):
:100:

PR #96, opened 2021-04-29
add strictNullChecks to tsconfig.json

3xtr4t3rr3str14l:
part of ChainSafe/lodestar#1830

PR #99, opened 2021-05-01
Typesafe library

dapplion:
- Use same Typescript and lint options as Lodestar monorepo
- Use same Typescript version as Lodestar monorepo

Those two changes have surfaced many issues which I've resolved as many as possiible. @wemeetagain I don't know how to handle the function calls that you do with null such as
https://github.com/ChainSafe/ssz/blob/36b30f8916a5911576b445e3eadf87639bb7e640/src/types/composite/abstract.ts#L103
please take a look and find a solution accepted by the Typescript compiler.

----

2. This PR also enforces correct types for ContainerType fields and ArrayType element type. See this test as an example of what compiles and what doesn't now
https://github.com/ChainSafe/ssz/blob/36b30f8916/test/unit/eth2Types.test.ts

3. Replaced the toHexString and fromHexString for simpler alternatives that are
  - toHexString: x12 faster (now 0.4 µs / op, 32 bytes input)
  - fromHexString: x2 faster (now 2.4 µs / op, 32 bytes input)


wemeetagain:
`struct_getSerializedLength` is called with `null` in cases where the type is "fixed-length" and not dependent on the value.
I think @3xtr4t3rr3str14l ran into this same issue in #96 

The issue is that the signature is different between fixed-length and variable-length types.
One naive approach could be to change the signature to `struct_getSerializedLength(value?: T): number`.
And just throw an error at runtime if the value isn't provided for variable-length types.

A different approach could be to cast `null`/`undefined` to `T` in the fixed-length type case.

wemeetagain:
> change the signature to struct_getSerializedLength(value?: T): number.

Seems we already did similar here in https://github.com/ChainSafe/ssz/pull/110 https://github.com/ChainSafe/ssz/pull/110/files#diff-f7dbd702062a823eb1050e5721c8802d7d613821a719d69da838bc0760cef64bR594-R596

Pushed a few commits on top to fix conflicts and type errors

dapplion:
conflicts

dapplion:
@wemeetagain Can you take a look?

dapplion:
PR got fucked with conflicts :sweat_smile: @wemeetagain do you think it's worth it to pursue this?

dapplion:
Implemented with https://github.com/ChainSafe/ssz/pull/223

Issue #100, opened 2021-05-01
toHexString performance issue


dapplion:
Running a Lodestar node in Prater with validators 3.25% of total CPU time is spent on `toHexString()`

![Screenshot from 2021-05-01 16-40-02](https://user-images.githubusercontent.com/35266934/116785855-062a7d00-aa9c-11eb-91f8-c6ce8cb6c559.png)

PR #101, opened 2021-05-01
Improve hexToString performance

dapplion:
See https://github.com/ChainSafe/ssz/issues/100 hexToString() is an significant performance bottleneck.

Replaced the toHexString and fromHexString for simpler alternatives that are

toHexString: x12 faster (now 0.4 µs / op, 32 bytes input)
fromHexString: x2 faster (now 2.4 µs / op, 32 bytes input)

Closes https://github.com/ChainSafe/ssz/issues/100

### DIFF
--- a/src/util/byteArray.ts
+++ b/src/util/byteArray.ts
@@ -1,19 +1,38 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
 import {ByteVector} from "../interface";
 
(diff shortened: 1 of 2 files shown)


3xtr4t3rr3str14l (approved):

dapplion:
@wemeetagain ping we may want to do a release just for this and add to Lodestar

wemeetagain (approved):

PR #110, opened 2021-05-19
Add composite type leaves to proof

wemeetagain:
When creating a proof with a path to a composite type, include all
leaves of the type in the proof.

- Add `CompositeType#tree_getLeafGindices(tree?: Tree, root?: Gindex): Gindex[]`
  - This is where the magic happens. Composite types are recursively traversed to the terminal types to get leaf gindices.
  - The tree is included for 'length' information for variable-length objects. (otherwise we wouldn't know which gindices to include for eg: `state.validators`)
  - A "root" gindex is included to allow for deeply nested types to return gindices relative to the root type
- `CompositeType#createProof` is extended to call `tree_getLeafGindices` when the type at a path is a composite type.

### DIFF
--- a/src/types/composite/abstract.ts
+++ b/src/types/composite/abstract.ts
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/member-ordering */
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import {CompositeValue, Json} from "../../interface";
-import {BackedValue, CompositeArrayTreeValue, createTreeBacked, isBackedValue, Path, TreeBacked} from "../../backings";
+import {BackedValue, createTreeBacked, isBackedValue, Path, TreeBacked} from "../../backings";
 import {IJsonOptions, isTypeOf, Type} from "../type";
 import {
   concatGindices,
--- a/src/types/composite/abstract.ts
+++ b/src/types/composite/abstract.ts
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/member-ordering */
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import {CompositeValue, Json} from "../../interface";
-import {BackedValue, createTreeBacked, isBackedValue, Path, TreeBacked} from "../../backings";
+import {BackedValue, CompositeArrayTreeValue, createTreeBacked, isBackedValue, Path, TreeBacked} from "../../backings";
 import {IJsonOptions, isTypeOf, Type} from "../type";
 import {
   concatGindices,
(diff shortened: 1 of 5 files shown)


src/types/composite/list.ts
```diff
@@ -376,4 +385,13 @@ export class CompositeListType<T extends List<object> = List<object>> extends Co
     this.tree_setLength(target, length - 1);
     return value as T[number];
   }
+  tree_getLeafGindices(target?: Tree, root: Gindex = BigInt(1)): Gindex[] {
+    if (!target) {
+      throw new Error("variable type requires tree argument to get leaves");
+    }
+    const gindices = super.tree_getLeafGindices(target, root);
+    // include the length chunk
+    gindices.push(concatGindices([root, BigInt(3)]));
```
dapplion, line 394: Why `3` here? Could you add a comment explaining

src/types/composite/array.ts
```diff
@@ -564,4 +583,31 @@ export abstract class CompositeArrayType<T extends ArrayLike<unknown>> extends C
...
+      for (
+        let i = 0, gindex = startIndex, extendedGindex = extendedStartIndex;
+        i < chunkCount;
+        i++, gindex++, extendedGindex++
+      ) {
+        gindices.push(...this.elementType.tree_getLeafGindices(target.getSubtree(gindex), extendedGindex));
+      }
+    } else {
+      for (let i = 0, gindex = extendedStartIndex; i < chunkCount; i++, gindex++) {
+        gindices.push(...this.elementType.tree_getLeafGindices(undefined, gindex));
```
dapplion, line 608: :+1: for push(... ) https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki

src/types/composite/list.ts
wemeetagain: The comment on the line above is an attempt to explain :sweat_smile:

dapplion (approved):
Looks great! :fire: Can't wait to deploy this to the lightclient demo site

src/types/composite/abstract.ts
```diff
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/member-ordering */
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import {CompositeValue, Json} from "../../interface";
-import {BackedValue, createTreeBacked, isBackedValue, Path, TreeBacked} from "../../backings";
+import {BackedValue, CompositeArrayTreeValue, createTreeBacked, isBackedValue, Path, TreeBacked} from "../../backings";
```
dapplion, line 5: `CompositeArrayTreeValue` is not used

dapplion (approved):

PR #124, opened 2021-06-08
Fix struct_getRootAtChunkIndex for BitVectorType

### LINKED ISSUE #123: Inconsistent hashTreeRoot() of BitVectorType
This unit test failed
```ts
it.only("SyncCommitteeBits", function () {
    const SyncCommitteeBits = new BitVectorType({
      length: 512,
    });
    const biStr = "00001110011100101010100110111001111011110111001110110010101000010010011110000110001101111100100100011011001001010000111010010011100100111010111101110110001000000011011001011000011101010111111011000110000101100111111000110011110010010110101011111110111010101111110010011111101001011110001101111110111001100110110001100010100010101110110010100100001011000101101000011010111010111000100100101000101100001100011001110100100111110011100111001100101001011011111001111010111011000100100000010000000111010010100000000111";
    const arr = Array.from({length: 512}, (_, i) => biStr.charAt(i) === "0" ? false: true);
    const rootByStruct = SyncCommitteeBits.hashTreeRoot(arr);
    const bytes = SyncCommitteeBits.serialize(arr);
    const rootByTreeBacked = SyncCommitteeBits.createTreeBackedFromBytes(bytes).hashTreeRoot();
    expect(toHexString(rootByStruct)).to.be.equal(toHexString(rootByTreeBacked), "Inconsistent hashTreeRoot");
  });
```
with
```
Inconsistent hashTreeRoot
      + expected - actual

      -0x293e7a2daf052e84155cbc618f5ab1873cac531310ea0ecbd5d6ac927797f3a5
      +0xdeda3ed719579278aaab269fbe7bec2ffa69f2d4d488d6ce50925424b6e7354c
      
      at Context.<anonymous> (test/unit/regression.test.ts:15:45)
      at processImmediate (internal/timers.js:461:21)
```

seems like the tree backed implementation was good while the struct hashTreeRoot() was not, see https://github.com/ChainSafe/lodestar/issues/2648

tuyennhv:
Close #123 
+ `BitVectorType.struct_getRootAtChunkIndex` accepts a chunk index, not byte index so we have to multiply 32 as byte offset
+ Rename variables accordingly and add unit test

### DIFF
--- a/src/types/composite/bitList.ts
+++ b/src/types/composite/bitList.ts
@@ -4,6 +4,7 @@ import {BasicListType} from "./list";
 import {booleanType} from "../basic";
 import {isTypeOf, Type} from "../type";
 import {fromHexString, toHexString} from "../../util/byteArray";
+import {BYTES_PER_CHUNK} from "../../util/constants";
 import {Tree} from "@chainsafe/persistent-merkle-tree";
 
 export interface IBitListOptions {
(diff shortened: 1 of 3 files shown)


dapplion (approved):

dapplion:
Nice find! Can you run Teku's devnet linking this fix to confirm it fixes the issue?

tuyennhv:
> Nice find! Can you run Teku's devnet linking this fix to confirm it fixes the issue?

just tried this, `Invalid state root` is still an issue.

dapplion:
> > Nice find! Can you run Teku's devnet linking this fix to confirm it fixes the issue?
> 
> just tried this, `Invalid state root` is still an issue.

Tho we should merge this anyway right?

dapplion (approved):

PR #145, opened 2021-08-13
Implement Union type

tuyennhv:
**Motivation**
+ This Union type was added in altair 1.1.0 alpha.7
+ Not used in phase0/altair
+ Don't see it in any spec tests

**Description**
+ New basic None type
+ New composite Union type. A Union type is configured using an array of type, None type (if exists) should be the 1st, for example
```
export const UnionObject = new UnionType({types: [new NoneType(), SimpleObject, number16Type]});
```
a value contains a selector (index of selected type) and an internal value. In the above example, if selector = 1 it's null, if selector = 1 it's a SimpleObject and if selector = 2 it's a number of 2 bytes
+ In a tree representation, selector is at gindex 3 which is similar to the length in an array


dapplion (dismissed):
Looks great! @wemeetagain can you take a look?

wemeetagain (dismissed):

wemeetagain:
file conflicts

dapplion (approved):

wemeetagain (approved):

PR #146, opened 2021-08-16
Cache field information for Container type

tuyennhv:
**Motivation**
+ For container type, we keep calculating gindex and `isCompositeType` for all fields

**Description**
+ Instead just calculate them once in the constructor and cache
+ The performance test `uint.test.ts` shows that it's 30% faster with this cache


src/types/composite/container.ts
```diff
@@ -26,16 +26,30 @@ export const CONTAINER_TYPE = Symbol.for("ssz/ContainerType");
...
+type FieldInfo = {
+  isBasic: boolean;
+  gindex: bigint;
+};
 
 export class ContainerType<T extends ObjectLike = ObjectLike> extends CompositeType<T> {
   // ES6 ensures key order is chronological
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   fields: Record<string, Type<any>>;
+  fieldInfos: Map<string, FieldInfo>;
```
dapplion, line 38: Can you add a nice JSDoc comment explaining the improvement with this cache, the expected speed up, etc?

dapplion (approved):

PR #159, opened 2021-08-19
Implement Number64UintArray and tree_newTreeFromUint64Deltas

### LINKED ISSUE #156: Consume HashObject in ssz and new array method
**Motivation**
1. HashObject

+ The new as-sha256 and persistent-merkle-tree implementation uses HashObject internally, we need to expose that
+ ssz would use it to probably extract/set value faster (for uint type) without having to use an intermediate Uint8Array

2. In order to update an array (list/vector) the consumer has to get array item, +/-delta and update back the item to the array
```ts
   state.balances[i] = state.balances[i] + delta;
```
we should be able to implement a new "`tree_`" method in the array like
```ts
  tree_updateUint64Delta(index: number, delta: number): number;
```
it'd save 1 time to access/traverse the tree.

tuyennhv:
**Motivation**
1. Since persistent-merkle-tree internally track HashObject, we want to use it directly instead of having intermediate Uint8Array
2. HashObject contains 8 32 bytes number which is suitable for serialize/deserialize uint
3. Instead of having to serialize/deserialize to apply a delta, like `balances[i] = balances[i] + delta` we want a method to apply delta directly to an item in `BasicArrayType`
4. New method to apply deltas and return a new tree, this is a convenient method for lodestar to process/apply rewards and penalties

**Description**
+ New `Number64UintType` for 64 bit uint, this is better than `NumberUintType({byteLength: 8}) - performance test proved that
+ `Number64UintType` works on HashObject instead of bytes, use it in `Container` and newly created `Number64ListType `
+ Number64ListType: new tree_applyUint64Delta method, this is for lodestar decreaseBalance increaseBalance scenario
+ Number64ListType: new `tree_newTreeFromUint64Deltas`, this is for lodestar `processRewardsAndPenalties` scenario

Closes #156

I applied this branch locally and got a better benchmark test result in lodestar (the result comes from different improvement, not only this PR)

|#|master|new branch|
|-|-------|------------|
|beforeProcessEpoch|1.603897  s/op|697.9248 ms/op|
|processRewardsAndPenalties|704.3871 ms/op|138.6842 ms/op|

Follow up work after this PR:
+ Use gindexbitstring whenever possible
+ Consider applying same strategy for BigIntUintType if bytelength = 8


package.json
```diff
@@ -30,8 +30,8 @@
   },
   "types": "lib/index.d.ts",
   "dependencies": {
-    "@chainsafe/as-sha256": "^0.2.2",
-    "@chainsafe/persistent-merkle-tree": "^0.3.3",
+    "@chainsafe/as-sha256": "^0.2.3",
+    "@chainsafe/persistent-merkle-tree": "../persistent-merkle-tree",
```
tuyennhv, line 34: waiting for https://github.com/ChainSafe/persistent-merkle-tree/pull/59

src/types/basic/uint.ts
```diff
@@ -123,6 +129,101 @@ export class NumberUintType extends UintType<number> {
...
+ * For 64 bit number, we want to operator on HashObject
+ * over bytes to improve performance.
+ */
+export class Number64UintType extends NumberUintType {
+  constructor() {
+    super({byteLength: 8});
+    this._typeSymbols.add(NUMBER_64_UINT_TYPE);
+  }
+
+  // TODO: move this logic all the way to persistent-merkle-tree
```
tuyennhv, line 144: right now for `tree_applyUint64Delta` scenario we have to:
+ traverse the tree based on gindex
+ deserialize, apply delta
+ serialize, traverse the tree to apply the result

technically we can move this logic all the way to persistent-merkle-tree to traverse the tree only 1 time, but is it weird to do that?

src/types/basic/uint.ts
```diff
@@ -128,6 +134,93 @@ export class NumberUintType extends UintType<number> {
...
+ * over bytes to improve performance.
+ */
+export class Number64UintType extends NumberUintType {
+  constructor() {
+    super({byteLength: 8});
+    this._typeSymbols.add(NUMBER_64_UINT_TYPE);
+  }
+
+  // TODO: move this logic all the way to persistent-merkle-tree?
+  struct_deserializeFromHashObject(data: HashObject, offset: number): number {
```
dapplion, line 150: Use chunkOffset or some name that clearly differentiates from regular byte offsets

src/types/composite/abstract.ts
```diff
@@ -166,6 +167,10 @@ export abstract class CompositeType<T extends CompositeValue> extends Type<T> {
     return target.getRoot(this.getGindexAtChunkIndex(index));
   }
 
+  tree_getHashObjectAtChunkIndex(target: Tree, index: number): HashObject {
```
dapplion, line 170: use chunkIndex as argument

src/types/composite/array.ts
```diff
@@ -229,12 +233,30 @@ export abstract class BasicArrayType<T extends ArrayLike<unknown>> extends Compo
   }
 
   tree_getValueAtIndex(target: Tree, index: number): T[number] {
+    // a special optimization for uint64
+    if (this.isNumber64UintType) {
```
wemeetagain, line 237: maybe you could make a subclass of BasicArrayType?

A `Number64Array`  that has an elementType of `Number64UintType`.

wemeetagain (changes requested):

src/types/basic/uint.ts
```diff
@@ -128,6 +134,96 @@ export class NumberUintType extends UintType<number> {
...
+    // a chunk contains 4 items
+    if (numberOffset < 0 || numberOffset > 3) {
+      throw new Error(`Invalid numberOffset ${numberOffset}`);
+    }
+    let low32Number = 0;
+    let high32Number = 0;
+
+    switch (numberOffset) {
+      case 0:
+        low32Number = data.h0 & 0xffffffff;
```
dapplion, line 163: Is it really necessary to mask with `0xffffffff`? Aren't we sure the number is already 32 bits?

src/util/hash.ts
```diff
@@ -1,9 +1,47 @@
...
+  h1: 0,
+  h2: 0,
+  h3: 0,
+  h4: 0,
+  h5: 0,
+  h6: 0,
+  h7: 0,
+};
+
+/** Clone a hash object using sharedHashObject */
```
dapplion, line 23: Could you add more details in the JSDoc comment about the mechanics of this sharedHashObject? It's not quite clear to me.

dapplion (commented):
Looks really good! Just some minor comments

src/types/basic/uint.ts
dapplion: Oh right, we 100% should only read each chunk once. Can you add a utility function to set deltas in batch such that each tree leaf chunk is read only once?

src/types/basic/uint.ts
tuyennhv: yeah it's not really necessary, it's just a pattern that make me feel safe and comfortable like everywhere else in ssz. It's a cheap operation anyway so I prefer keeping it.

src/util/hash.ts
tuyennhv: this is used in a lot of places then we apply the shared HashObject to persistent-merkle-tree, it'll make a copy so it's safe to mutate it after that.
so yes, it will get mutated in different places but it's safe to do so.
I'll add more comments regarding this.

src/types/basic/uint.ts
tuyennhv: > Even at the epoch transition we will likely recreate the entire tree. So can we have a fast utility to do that with minimal tree traversing and touching each chunk exactly once?

for epoch transition, I already traverse each chunk once by leveraging `getNodesAtDepth()` api, please have a look at `number64_newTreeFromUint64Deltas()` implementation.

My concern was mainly for the `tree_applyUint64Delta` flow, in lodestar it'd improve the `processBlock()` flow where we have to apply deltas for sync committee balances.

Initially I thought it's weird having to bring this logic to persistent-merkle-tree, but we could extend the Tree class and implement it right in ssz, and use the new Tree here.

> Oh right, we 100% should only read each chunk once. Can you add a utility function to set deltas in batch such that each tree leaf chunk is read only once?

👍  I'd love to implement that in a separate PR

dapplion (approved):
:fire:

dapplion (dismissed):

wemeetagain (approved):
I pushed a commit on top, let me know what you think. I'd like to add methods to serialize/deserialize from HashObject for all BasicType s eventually (Mainly will be applicable to BooleanType). Ideally, the HashObject path should be taken when eg: `validators[i].slashed` is checked or set.

Here's a list of things changed:
Add optional tree_* hash object methods to BasicType - this lets us check for the existence of the method instead of using FieldInfo.isNumber64Type, useful for the scenario above
Remove FieldInfo.isNumber64Type
Remove separate number64_* functions - we can always split these out if we use them in multiple places
Rename applyDeltas methods - remove the word Uint64 from these methods

tuyennhv:
> I pushed a commit on top, let me know what you think. I'd like to add methods to serialize/deserialize from HashObject for all BasicType s eventually (Mainly will be applicable to BooleanType). Ideally, the HashObject path should be taken when eg: `validators[i].slashed` is checked or set.
> 
> Here's a list of things changed:
> 
> * Add optional tree_* hash object methods to BasicType - this lets us check for the existence of the method instead of using FieldInfo.isNumber64Type, useful for the scenario above
> * Remove FieldInfo.isNumber64Type
> * Remove separate number64_* functions - we can always split these out if we use them in multiple places
> * Rename applyDeltas methods - remove the word Uint64 from these methods

Thanks @wemeetagain for all of this 👍

PR #167, opened 2021-08-20
Avoid iterateNodesAtDepth

wemeetagain:
Use `getNodesAtDepth` instead of `iterateNodesAtDepth`.
Update persistent-merkle-tree

### DIFF
--- a/src/types/composite/array.ts
+++ b/src/types/composite/array.ts
@@ -173,7 +173,7 @@ export abstract class BasicArrayType<T extends ArrayLike<unknown>> extends Compo
     const remainder = size % 32;
     let i = 0;
     if (fullChunkCount > 0) {
-      for (const node of target.iterateNodesAtDepth(this.getChunkDepth(), 0, fullChunkCount)) {
+      for (const node of target.getNodesAtDepth(this.getChunkDepth(), 0, fullChunkCount)) {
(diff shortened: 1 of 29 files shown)


tuyennhv (approved):

tuyennhv:
there is lint error!

src/types/composite/array.ts
```diff
@@ -173,7 +173,7 @@ export abstract class BasicArrayType<T extends ArrayLike<unknown>> extends Compo
     const remainder = size % 32;
     let i = 0;
     if (fullChunkCount > 0) {
-      for (const node of target.iterateNodesAtDepth(this.getChunkDepth(), 0, fullChunkCount)) {
+      for (const node of target.getNodesAtDepth(this.getChunkDepth(), 0, fullChunkCount)) {
```
dapplion, line 176: Can we benchmark using a regular `for (;;)` loop instead of `for of`? If we are removing iterators, we should go all the way.

wemeetagain: Ok, I'll remove `for of` in consumption of `getNodesAtDepth`. Further use of iterators should be evaluated separately since that will likely require slightly deeper code changes. Preference for keeping the PR light and not expanding the scope if possible.

dapplion (approved):

PR #168, opened 2021-08-20
Add ContainerLeafNodeStructType for memory efficiency

dapplion:
Adds a new type `ContainerLeafNodeStructType` which represent its underlying data as struct when tree backed.

This special type comes from the need in Lodestar beacon node to represent the validators registry in a memory efficient way that allows for a fast state transition function. The validator registry is read often but written to very rarely, so this trade-off is really good.

The `ContainerLeafNodeStructType` requires other special structures to work:
- `BranchNodeStruct`: Custom persistent-merkle-tree node that holds the struct value and merkelizes on demand to hash and compute proofs
- `ContainerLeafNodeStructTreeValue`: Custom tree value that returns struct values


src/types/composite/container.ts
```diff
@@ -533,3 +535,106 @@ export class ContainerType<T extends ObjectLike = ObjectLike> extends CompositeT
...
+    return yield* this.tree_iterateValues(target);
+  }
+}
+
+export class BranchNodeHybrid<T> extends Node {
+  constructor(readonly type: ContainerTypeHybrid<T>, readonly value: T) {
+    super();
+  }
+
+  get root(): Uint8Array {
```
tuyennhv, line 618: it looks like `left()` and `right()` expect the value can be changed so it always have to recalculate the tree while this `root()` getter try to cache the root.
Also the `_root` has gone in `BranchNode` due to the latest persistent-merkle-tree, let's use h0 condition instead https://github.com/ChainSafe/persistent-merkle-tree/blob/d9d35f8fc373a0e76307841c8d27e42de252cee9/src/node.ts#L45

dapplion: Updated to use new hashObject. left and right don't cache since that would defeat the purpose of memory efficiency

test/unit/validators.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    activationEligibilityEpoch: 1_000_000,
+    activationEpoch: 2_000_000,
+    exitEpoch: 3_000_000,
+    withdrawableEpoch: 4_000_000,
+  };
+
+  const validatorTb = ValidatorType.createTreeBackedFromStruct(validator);
+  const validatorTbLeafNodeStruct = ValidatorLeafNodeStructType.createTreeBackedFromStruct(validator);
+
+  const ops: Record<string, (treeBacked: TreeBacked<Validator>, type: ContainerType<Validator>) => unknown> = {
```
tuyennhv, line 61: may not need `type` param here.

also, I'd like to add a test to mutate any fields and compare all ops again

src/types/composite/containerLeafNodeStruct.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  /** Overrides to return BranchNodeStruct instead of regular Tree */
+
+  createTreeBacked(tree: Tree): TreeBacked<T> {
+    const value = new ContainerLeafNodeStructTreeValue(this, tree);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return (new Proxy(value, TreeProxyHandler as any) as unknown) as TreeBacked<T>;
+  }
+
+  // struct_defaultValue
```
tuyennhv, line 29: need to clean up those

test/unit/validators.test.ts
tuyennhv: also, we want to make sure it works in a parent tree, ie create a validator list, modify some properties at some indices and compare the parent root (validator list root) again

tuyennhv (approved):

tuyennhv (approved):

PR #171, opened 2021-08-24
Add alternative to iterator interface

wemeetagain:
Update persistent-merkle-tree to v0.3.6 - this makes `tree_getValue` much faster, roughly 2x increase
Add `tree_getValues` as a non-iterator alternative to `tree_iterateValues`
Add `tree_readonlyGetValue`  as a non-iterator alternative to `tree_readonlyGetValues`
Expose non-iterator alternatives to `keys`/`values`/`entries` on `TreeBacked`/`TreeValue`

### DIFF
--- a/test/perf/iterate.test.ts
+++ b/test/perf/iterate.test.ts
@@ -1,4 +1,14 @@
 import {itBench, setBenchOpts} from "@dapplion/benchmark";
+import {
+  booleanType,
+  ByteVectorType,
+  ContainerType,
+  List,
+  ListType,
+  NumberUintType,
+  ObjectLike,
+  TreeBacked,
+} from "../../src";
 
 describe("iterate", () => {
   setBenchOpts({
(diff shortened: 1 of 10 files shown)


test/perf/iterate.test.ts
```diff
@@ -27,3 +37,82 @@ describe("iterate", () => {
...
+  itBench("basicListValue.readonlyValues()", () => {
+    for (const balance of balances.readonlyValues()) {
+      balance + 1;
+    }
+  });
+  itBench("basicListValue.readonlyValuesArray()", () => {
+    const balancesArray = balances.readonlyValuesArray();
+    for (let i = 0; i < balancesArray.length; i++) {
+      balancesArray[i] + 1;
+    }
```
dapplion, line 60: This comparision is not totally right, since for the array version it iterates twice. A more proper would be
```ts
  itBench("basicListValue.readonlyValues()", () => {
    Array.from(balances.readonlyValues())
  });
  itBench("basicListValue.readonlyValuesArray()", () => {
    balances.readonlyValuesArray();
  })
```

PR #176, opened 2021-08-30
dataview based readwrite num bytes opts

### LINKED ISSUE #143: Fast conversion from decimal to bytes
Lodestar spends significant CPU time converting uint types from decimal to bytes and back. For example, converting the balance tree to flat and back consist of doing 200_000 ops of those.

SSZ is the one hosting the code that does the conversions. This is one of the implementations that participate in this conversions.

https://github.com/ChainSafe/ssz/blob/be03a3398e9c13b4d6c609ba07dfba7bf2cfe2a6/src/types/basic/uint.ts#L68-L82

https://github.com/ChainSafe/ssz/blob/be03a3398e9c13b4d6c609ba07dfba7bf2cfe2a6/src/types/basic/uint.ts#L84-L98

We should investigate faster alternatives

g11tech:
This PR replaces the byte extract from uint number via native dataview, which is roughly 2x faster extraction than normal loop. 
when plugged in it gives whopping 80% performace improvement over normal uint byte extraction and a marginal 3-5% improvement in the new uint64 tree backed datatype manipulation.

before: 
image

after:
image

partially closes #143

### DIFF
--- a/src/types/basic/uint.ts
+++ b/src/types/basic/uint.ts
@@ -3,6 +3,12 @@ import {bigIntPow} from "../../util/bigInt";
 import {isTypeOf, Type} from "../type";
 import {BasicType} from "./abstract";
 import {HashObject} from "@chainsafe/as-sha256";
+import {
+  getNumberFromBytesLE,
+  getBytesFromNumberLE,
+  getNumberFromBytesLE32,
+  getBytesFromNumberLE32,
+} from "../../util/numhelper";
 
 export interface IUintOptions {
   byteLength: number;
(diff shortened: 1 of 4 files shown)


dapplion:
Yeah DataView are so much faster than manual manipulation with currently do. All this optimizations are found in a similar form in https://github.com/ChainSafe/ssz/pull/223 should we close this one in favor of https://github.com/ChainSafe/ssz/pull/223 ?

wemeetagain:
@dapplion :+1:

PR #179, opened 2021-08-30
Fix ContainerLeafNodeStructTreeValue

dapplion:
With current master the validator object is returned as `ContainerTreeValue`. However since it's not really tree backed acessing it's properties errors with 

```
  1) Container with BranchNodeStruct
       ValidatorLeafNodeStructType in List
         Should return pubkey as struct backed:
     TypeError: target.getNodesAtDepth is not a function
      at ByteVectorType.tree_convertToStruct (src/types/composite/byteVector.ts:59:27)
      at Proxy.valueOf (src/backings/tree/treeValue.ts:125:22)
      at Context.<anonymous> (test/unit/validators.test.ts:127:52)
      at processImmediate (internal/timers.js:464:21)

```

This PR makes createTreeBacked return a TreeValue that can deal with LeafNodeStruct nodes

### DIFF
--- a/src/backings/tree/treeValue.ts
+++ b/src/backings/tree/treeValue.ts
@@ -14,6 +14,7 @@ import {
   isVectorType,
   isUnionType,
   UnionType,
+  isContainerLeafNodeStructType,
 } from "../../types";
 import {byteArrayEquals} from "../../util/byteArray";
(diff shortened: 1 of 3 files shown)


wemeetagain (commented):
:+1:

PR #184, opened 2021-09-01
extending the case transformation in container data fill

g11tech:
extending the case transformation to other cases available in the "case" library in container data fill plus adding a `notransform` option and making no transformation default
plus test case for `notranform`

### DIFF
--- a/packages/ssz/src/types/composite/container.ts
+++ b/packages/ssz/src/types/composite/container.ts
@@ -248,7 +248,8 @@ export class ContainerType<T extends ObjectLike = ObjectLike> extends CompositeT
     const value = {} as T;
     for (const [fieldName, fieldType] of Object.entries(this.fields)) {
       const expectedCase = options && options.case;
-      const expectedFieldName = toExpectedCase(fieldName, expectedCase);
+      const customCasingMap = options && options.casingMap;
+      const expectedFieldName = toExpectedCase(fieldName, expectedCase, customCasingMap);
       if ((data as Record<string, Json>)[expectedFieldName] === undefined) {
         throw new Error(`Invalid JSON container field: expected field ${expectedFieldName} is undefined`);
       }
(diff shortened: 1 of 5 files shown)


dapplion:
@g11tech Can you make sure that the casing `signedHeader1` is correctly converted to `signed_header_1`? https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#proposerslashing

g11tech:
> @g11tech Can you make sure that the casing `signedHeader1` is correctly converted to `signed_header_1`? https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#proposerslashing

Ok 👍

g11tech:
comment: @dapplion  does this looks like correct parsing to you? (it will easily cover the signed_header_1 case):

`amd122cSigned3HeaderOpts1223` : `amd_122c_signed_3_header_opts_1223`

dapplion:
I got this error while fetching a block with attester slashing. The casing is definitely not correct

```
Error: Invalid JSON container field: expected field attestation1 is undefined
    at ContainerType.struct_convertFromJson (/home/lion/Code/eth2.0/lodestar/node_modules/@chainsafe/ssz/src/types/composite/container.ts:250:15)
```

g11tech:
@dapplion modified Case vs a new library "change-case" for snake conversion while dealing with numbers (yes both of them will easily cover our spec needs):

image

new library "change-case" performs marginally better (and also has like 2m weekly downloads) compared to current "case" (180k weekly downloads) and also has more extensible api than current one.

g11tech:
just noticed that there is also eth1Data that should be mapped to eth1_data (rather than eth_1_data .  I can however figure out a regex which can have the number chunkify rule to match eth1 together as a single unit (and supply that to the case library splitter)

dapplion:
> @dapplion modified Case vs a new library "change-case" for snake conversion while dealing with numbers (yes both of them will easily cover our spec needs):
> 
> image
> 
> new library "change-case" performs marginally better (and also has like 2m weekly downloads) compared to current "case" (180k weekly downloads) and also has more extensible api than current one.

Let's switch then :+1:

packages/ssz/src/types/composite/container.ts
```diff
@@ -248,7 +248,8 @@ export class ContainerType<T extends ObjectLike = ObjectLike> extends CompositeT
     const value = {} as T;
     for (const [fieldName, fieldType] of Object.entries(this.fields)) {
       const expectedCase = options && options.case;
-      const expectedFieldName = toExpectedCase(fieldName, expectedCase);
+      const customCasingMap = options && options.casingMap;
+      const expectedFieldName = toExpectedCase(fieldName, expectedCase, customCasingMap);
```
dapplion, line 252: Should we add this json options as an option to the container itself too? That would make declaration easier such that you define the fields and the casing in one place

g11tech: i am in favor of that, especially when there is container nesting (and it will be more optimized as the lookup will be in the container fields local scope) . you want to do it in this PR or separate?

dapplion: Good point, do it a another one :+1:

dapplion (approved):
Nice work!

PR #186, opened 2021-09-07
Combine ssz-adjacent libraries into a lerna monorepo

wemeetagain:
Combine the following repos into a lerna monorepo:
- as-sha256
- persistent-merkle-tree
- persistent-ts
- ssz
- simpleserialize.com


dapplion:
:fire: yeah! Looking great!

Any clue why so many performance tests are so much slower in this PR?

tuyennhv:
@wemeetagain there's a performance issue with the way tsc 4.4.2 compiles persistent-merkle-tree `node.ts` file, we should track it separately, I created #187

I just limited typescript in this PR to 4.2.x and it works fine.

wemeetagain:
@ChainSafe/eth2-0 This should be ready for review

tuyennhv (approved):

Issue #195, opened 2021-09-14
statically providing casingMap while creating a new container type.


g11tech:
While figuring out a way to provide case transformation in the container datafill (toJson/fromJson calls), it was brainstormed that other elegant solution will be a way provide an optional casingMap while declaring the container (which can not only be used to case transform custom fields but as well as as speedup the normal fields lookup) and will also work well in the nested containers scenario having different casing of the fields.

**Is your feature request related to a problem? Please describe.**
case transformation of fields while filling/extracting container data from/to json.

**Describe the solution you'd like**
Declaring the casing structure of the fields while we create a container type.

**Describe alternatives you've considered**

**Additional context**
![image](https://user-images.githubusercontent.com/76567250/133231359-947c7741-0670-4345-9370-860cdefde488.png)

g11tech:
@dapplion @wemeetagain  how does this example  looks for providing the declaration time casingMap
```
const SlashingTransformFieldObject = new ContainerType({
  fields: {
    eth1Data: new NumberUintType({byteLength: 4}),
    signedHeader1: new NumberUintType({byteLength: 4}),
    signedHeader2: new NumberUintType({byteLength: 4}),
    attestation1: new BooleanType(),
    attestation2: new BooleanType(),
  },
  casingMap: {
    eth1Data: {default:"camel", snake: "eth1_data", ...},
    attestation1: {snake: "attestation_1"} 
  }
});
```
if there is a snake conversion (or default snake) provided and a casing has to be derived (not present in the casingMap), it can be used to chunkify and transform correctly.

also another question pops to mind, if a dynamic casing happens via toJson/fromJson, should we allow a flag to cache it so that next time around it just picks from the cache?

dapplion:
@g11tech I would keep it simple

```ts
const SlashingTransformFieldObject = new ContainerType({
  fields: {
    eth1Data: new NumberUintType({byteLength: 4}),
    signedHeader1: new NumberUintType({byteLength: 4}),
    signedHeader2: new NumberUintType({byteLength: 4}),
    attestation1: new BooleanType(),
    attestation2: new BooleanType(),
  },
  casingMap: {
    eth1Data: "eth1_data",
    attestation1: "attestation_1"
    ...
  }
});
```

Casing map provides the "correct" casing regardless of the name. If that's provided that's what has to be used ignoring any other option

g11tech:
[...]

PR #198, opened 2021-09-19
declaration time casingMap provisioning, and if declared always use the map first

g11tech:
**Motivation**

This PR enables one to provide a declared casing during creation of a container type. This will help us do declare direct match between expected api data to the container fields even when there is container nesting.
for fields which are not provided in the `casingMap` container declaration time, it will do no transformation assuming their casing is correct w.r.t to the json data. 
`notranform` case type has been repurposed and renamed to `declared`
This casingMap can be extended in future to make the UX more generic and smart.

example (also present in test case)
```typescript
export const WithCasingDeclarationFieldObject = new ContainerType({
  fields: {
    eth1Data: new NumberUintType({byteLength: 4}),
    signedHeader1: new NumberUintType({byteLength: 4}),
    signedHeader2: new NumberUintType({byteLength: 4}),
    attestation1: new BooleanType(),
    attestation2: new BooleanType(),
  },
  casingMap: {
    signedHeader1: "signed_header_1",
    signedHeader2: "signed_header_2",
    attestation1: "attestation_1",
    attestation2: "attestation_2",
  },
});
```

use:
```typescript
 const jsonData = {eth1_data: 11, signed_header_1: 4, signed_header_2: 5, attestation_1: true, attestation_2: false};
 const containerObject = WithCasingDeclarationFieldObject.fromJson(jsonData, {
    case: "snake",
  });
```
**Description**

Closes #issue_number
#195

### DIFF
--- a/packages/ssz/src/types/composite/container.ts
+++ b/packages/ssz/src/types/composite/container.ts
@@ -250,7 +250,7 @@ export class ContainerType<T extends ObjectLike = ObjectLike> extends CompositeT
     }
     const value = {} as T;
     const expectedCase = options && options.case;
-    const customCasingMap = expectedCase === "declared" ? this.casingMap : options && options.casingMap;
+    const customCasingMap = this.casingMap || (options && options.casingMap);
     for (const [fieldName, fieldType] of Object.entries(this.fields)) {
       const expectedFieldName = toExpectedCase(fieldName, expectedCase, customCasingMap);
       if ((data as Record<string, Json>)[expectedFieldName] === undefined) {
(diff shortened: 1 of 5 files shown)


packages/ssz/src/types/composite/container.ts
```diff
@@ -246,9 +249,9 @@ export class ContainerType<T extends ObjectLike = ObjectLike> extends CompositeT
       throw new Error("Invalid JSON container: expected Object");
     }
     const value = {} as T;
+    const expectedCase = options && options.case;
+    const customCasingMap = expectedCase === "declared" ? this.casingMap : options && options.casingMap;
```
dapplion, line 253: if `this.casingMap` exists, then ignore all json options and always apply that mapping

packages/ssz/src/util/json.ts
```diff
@@ -8,7 +8,7 @@ export function toExpectedCase(
 ): string {
   if (customCasingMap && customCasingMap[value]) return customCasingMap[value];
   switch (expectedCase) {
-    case "notransform":
+    case "declared":
```
dapplion, line 11: Remove

g11tech: this is when the field is not present in the casingMap

dapplion (approved):
Looks great!

PR #201, opened 2021-09-25
ssz version 0.8.18 release

g11tech:
## Features

- Ability to specify casingMap declaration time for Container's toJson/fromJson [#198](https://github.com/ChainSafe/ssz/pull/198)
- Extending case matching in Container's toJson/fromJson with a range of case types [#184](https://github.com/ChainSafe/ssz/pull/184)
- Ability to provide casingMap in toJson/fromJson interface for aiding case matching [#184](https://github.com/ChainSafe/ssz/pull/184)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,9 +6,33 @@ on:
       - "master"
 
 jobs:
+  tag:
+    name: Check and Tag
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Get latest tag
+        id: get-latest-tag
+        uses: actions-ecosystem/action-get-latest-tag@v1
+        with:
+          with_initial_version: false
+      - name: Create tag
+        id: tag
+        uses: butlerlogic/action-autotag@1.1.2
+        with:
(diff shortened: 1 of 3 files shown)


g11tech:
@dapplion  i noticed that the current release script doesn't has the tagging section.
should that be added and release continue with tag 0.8.18 since it would be first release post monorepo merge

dapplion:
@g11tech Indeed the release workflow is missing stuff, could you improve it to mimic the one we have in Lodestar that also creates Github releases?

dapplion (approved):

g11tech:
@dapplion the release workflow is failing on chainsafe/ssz because of changelog generation step, and after trying to dig a little figured out, the image is sort of not able to not able to map pre mono repo ssz tags with current repo and throwing error of `common ancestor not found`. 
in my own github account, i deleted all pre mono repo ssz tags, then was able to successfully go past this and sort of create a dummy release .

should i delete the pre mono repo ssz tags on the main chainsafe/ssz branch?

dapplion:
@g11tech Right I was wrong to just use the same workflow as Lodestar because here the versioning is independent. Please coordinate with @wemeetagain to develop merge and test a workflow for independent versioning releases

Issue #203, opened 2021-09-29
mono repo release workflow


g11tech:
the current release workflow is failing post monorepo structure change.

1. changelog step is giving issue with the pre monorepo tags created - suggestion delete them (verified in my  forked clone of account)  https://github.com/ChainSafe/ssz/pull/201#issuecomment-930390619

2. release version number is being picked from packages/ssz/package.json, its currently 0.8.18, but there are tags imported from other repos as consequence of merge like 0.19... So should we do the tagging with 0.20. something?
3. since the other packages are on different versions, should be track the release version and tag from the main package.json? that version will be required to be incremented each time a child package needs publishing. in lodestar we update versions of all packages

g11tech:
@dapplion @wemeetagain

wemeetagain:
Yeah, the main difference between the lodestar monorepo and this is that ssz uses independent versioning.
So likely tagging, changelog management, github releases will have to managed differently than lodestar.

Tags will probably look like: ${packageName}@${packageVersion}
Changelogs will either need to be separated by package name in a single file or separated by package directory
Github releases should probably happen for each package release (or maybe just have a github release for the ssz package).

I think ideally all versioning/changelog/github releasing can be handled by lerna. `lerna version` to increment versions + git tag + generate changelog as needed. `lerna publish` to publish versions to npm.

g11tech:
@wemeetagain lerna publish seemed to version and push to git (origin) as well as i ran `yarn lerna publish patch    --skip-npm` and it created and published tags and changelog in my account's forked ssz repo.
(`yarn lerna publish from-package   --skip-npm` was publishing null versions because i guess skip npm was not allowing it to fetch versions)

![image](https://user-images.githubusercontent.com/76567250/135426242-7035877c-a9e1-4b00-93d2-8f2280b2b384.png)

lerna publish seems to be doing bot `lerna version` and publishing, tagging pushing tags and creating release

g11tech:
also doing a full test release on my account's forked repo, the lerna seems to be asking to configure user.email, user.name:

![image](https://user-images.githubusercontent.com/76567250/135449752-ccb11e03-a864-4df0-afe2-4b23267f531d.png)

once i also added that in my repo's release,yml, things finally worked out (with `lerna publish patch --yes --no-verify-access --skip-npm`)

![image](https://user-images.githubusercontent.com/76567250/135449999-d556d7b4-7ff6-4e80-a709-e6a85c03559a.png)

[...]

PR #204, opened 2021-09-30
lerna publish based release workflow

g11tech:
**Motivation**

reverting the changes made to release.yml  in favour of  lerna publish based release workflow which seems to be tagging, creating reease, changelog as well apart from publishing to the npm repos.

This workflow has been tested in the @g11tech/ssz forked repo using `lerna publish patch --yes --no-verify-access --skip-npm` publish command instead of the original `lerna publish from-package --yes --no-verify-access` for testing all aspects of the workflow (except fetching versions and deploying to npm)

attached is the successful run in the forked repo:
image

**Description**

Closes #issue_number
#203

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,10 +22,14 @@ jobs:
       - run: yarn install --frozen-lockfile --ignore-optional && yarn build
       - run: yarn build
 
+      - run: git config user.email $(git --no-pager show -s --format='%ae' HEAD)
+      - run: git config user.name $(git --no-pager show -s --format='%an' HEAD)
+
       - name: Publish to npm registry
         run: yarn publish:release
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Deploy Website
         uses: peaceiris/actions-gh-pages@v2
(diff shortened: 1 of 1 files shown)


g11tech:
@wemeetagain have played around with lerna publish quite a bit in my home forked repo, in the current format it should work using "from-package", and generate changelog as well for the packages.
should we try releasing with this 0.8.18 version is already updated in ssz/package.json? (which is what we originally wanted to publish)

wemeetagain:
This looks good! We'll see how it runs.

I think the from-package option will be helpful for us so we may control which packages to release/publish. The process may be to make a PR where we run `lerna version`, which lets us interactively increment the versions of relevant packages. Then once merged, the `lerna publish` github workflow can publish to npm and github releases.

wemeetagain (approved):

g11tech:
> This looks good! We'll see how it runs.
> 
> I think the from-package option will be helpful for us so we may control which packages to release/publish. The process may be to make a PR where we run `lerna version`, which lets us interactively increment the versions of relevant packages. Then once merged, the `lerna publish` github workflow can publish to npm and github releases.

yea prepare: lerna version and commit
github-actions: release on merge


There is one step that we might need to add : early test and bail if no version change, right now all this comes to post install/build step (wasn't working when i moved it up), so will later figure it out if this works :+1:

PR #212, opened 2021-11-21
Add root readme

wemeetagain:
**Motivation**

Currently, there isn't a root README, makes the link https://github.com/chainsafe/ssz uninviting

**Description**

Add a readme linking to sub-packages and linking back to https://lodestar.chainsafe.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,19 @@
+# @Chainsafe/ssz
+
+> Typescript Simple Serialize (SSZ) libraries
+
+## Intro
+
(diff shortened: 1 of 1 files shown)


tuyennhv:
looks like we have to improve the performance test, updating README caused some performance regression

wemeetagain:
Pretty sure its just a bad run. No way for a readme to affect performance, right?

dapplion (approved):

PR #223, opened 2022-01-14
SSZ v2

dapplion:
**Motivation**

Our current library has multiple fundamental limitations that affect performance, memory and usability downstream. In Lodestar we have a added a lot of custom code to get around this limitations. This PR up-streams all the good ideas created downstream into a library that's performant by default without requiring extra infrastructure

### TreeViewDU

Stands for **TreeView Deferred Update**, and allows for very fast reads and writes without committing to the tree until latter. Then is optimized batched operations all nodes are recomputed and added to the tree with a fast routine `setNodesAtDepth`.

This mode is meant to be used in the entire beacon state transition function. This is a fundamental shift from the previous version where tree views and regular values had the same API. However, achieving this requires ES6 Proxy which have performance issues. Now the beacon state transition function will not compile if regular JS values are passed to it, it will only accept the TreeViewDU representation of the state.

```ts
const statuses = state.currentEpochParticipation;
for (const [index, delta] of deltas) {
  statuses.set(index, statuses.get(index) + delta)
}
state.commit()
```

### DataView and Uint8Array for faster (de)serialization

Current version spends significant time converting bytes to decimal numbers. [DataView](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) is extremely fast at bytes to decimal conversion and back. This PR uses it for value conversions and to populate HashObjects much faster.

However, `Uint8Array.slice` is by far the fastest way of copying binary data. So this PR passes both a DataView and a Uint8Array to make decimal conversions and data copying as fast as possible.

### Typed types

The Typescript types of values, TreeView and TreeViewDU are derived from the type arguments. Previously the SSZ type and its Typescript type had no assurance that they were correct and represented the same data. With this guarantee we can now use novel type representations and iterate quickly without breaking any code downstream.

```ts
const containerType =  new ContainerType({a: new UintNumberType(8)});
const container = containerType.defaultValue; // automatically typed as {a: number}
```

### Value hash cache

Previously only TreeViews had the ability to cache hashTreeRoot results, so in Lodestar we deserialized Attestations as tree just for that. This version introduces the option `cachePermanentRootStruct` for composite types which will cache only the root hashTreeRoot (not other intermediary hashes) in the JavaScript object itself. Now we can deserialize Attestations as value and get cached hashTreeRoot in a memory efficient and performant representation.

```ts
new ContainerType(fields, {cachePermanentRootStruct: true})
```

### BitArray

BitList and BitVectors are now represented in a new class `BitArray` that replaces the previous `boolean[]` representation. Thanks to some optimizations introduced in Lodestar to intersect Uint8Arrays fast it provides the same performance as a `boolean[]` but at a much lower memory cost. Now in Lodestar we don't need to do the optimization manually and just call

```ts
BitArray.intersectValues(committeeIndices);
```

### Pre-computed type data

Many of the types immutable characteristics like minSize, fixedLen, offset positions, etc are now computed in the constructor. This makes consuming this data instant speeding up serialization + de-serialization. Containers also pre-compute offset data and JSON keys data, which help speed up its methods.

```ts
type.minSize;
```

**TODO after merge**

- [ ] Make subtreeFillToContents not input data, and more benchmarks
- [ ] Add coverage tracker in SSZ
- [ ] Optimize merkelize chunks for hashTreeRoot
- [ ] Optimize hashing of struct values, 30-40% potential improvements to be gained. Matters when hashing the state for the first time.
- [ ] Optimize ArrayBasicView getAll()
- [ ] Optimize and benchmark setNodeAtDepth. Benchmark vs caching gindexstring


packages/ssz/src/type/vectorBasic.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  // Helpers for TreeView
+
+  tree_getLength(): number {
+    return this.length;
+  }
+
+  tree_setLength(): void {
+    //
```
tuyennhv, line 117: add comments why we don't need to implement this for Vector

packages/ssz/src/type/vectorBasic.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  tree_getChunksNode(node: Node): Node {
+    return node;
+  }
+
+  tree_setChunksNode(rootNode: Node, chunksNode: Node): Node {
+    return chunksNode;
```
tuyennhv, line 125: similar to the above, add a comment here

packages/ssz/src/view/arrayBasic.ts
```diff
@@ -0,0 +1,97 @@
...
+  TreeView<ArrayBasicType<ElementType>>,
+  TreeViewDU<ArrayBasicType<ElementType>>
+> & {
+  readonly elementType: ElementType;
+  readonly itemsPerChunk: number;
+  readonly chunkDepth: number;
+  tree_getLength(node: Node): number;
+  tree_setLength(tree: Tree, length: number): void;
+
+  // TEMP
```
tuyennhv, line 16: remove this or drop some comments

packages/ssz/test/unit/byType/listBasic/tree.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      // push 2 times to increase depth and test the expand functionality
+      id: "push basic x2",
+      valueBefore: [],
+      valueAfter: [1, 2],
+      fn: (tv) => {
+        tv.push(1);
+        tv.push(2);
+      },
+    },
+    {
```
tuyennhv, line 28: in this test, byteLength = 8, with a chunk of 32 bytes => itemsPerChunk = 4, we should add a test to push 5 times to add 1 more chunk

packages/ssz/src/viewDU/container.ts
```diff
@@ -0,0 +1,222 @@
...
+          this.type.fieldsEntries[index].fieldType as unknown as CompositeType<unknown, unknown, unknown>
+        ).commitViewDU(view),
+      });
+    }
+
+    for (const index of this.nodesChanged) {
+      nodesChanged.push({index, node: this.nodes[index]});
+    }
+
+    // TODO: Optimize to loop only once
```
tuyennhv, line 89: implement this TODO ?

wemeetagain (changes requested):

packages/persistent-merkle-tree/src/tree.ts
```diff
@@ -369,3 +189,352 @@ export class Tree {
...
+  // Optimized paths for short trees (x20 times faster)
+  if (depth === 0) {
+    return startIndex === 0 && count > 0 ? [rootNode] : [];
+  } else if (depth === 1) {
+    if (count === 0) {
+      return [];
+    } else if (count === 1) {
+      return startIndex === 0 ? [rootNode.left] : [rootNode.right];
+    } else {
+      return [rootNode.left, rootNode.right];
```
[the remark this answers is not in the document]
dapplion: Yeah this function is optimistic, as well as setNodesAtDepth where you must pass data sorted, and perfectly arranged, or the result is wrong in hard to notice ways. This methods are meant to achieve the max performance possible. getNodesAtDepth is a pretty hot path when deserializing a full state. Do you really think it should protect against the test cases you listed above?

packages/ssz/src/README.md
```diff
@@ -0,0 +1,85 @@
...
+- Attester objects from gossip must be deserialized fast to struct and serialized at ok speeds for broadcasting
+
+```
+ ___      ___  _______
+|\  \    /  /|/  ___  \
+\ \  \  /  / /__/|_/  /|
+ \ \  \/  / /|__|//  / /
+  \ \    / /     /  /_/__
+   \ \__/ /     |\________\
+    \|__|/       \|_______|
```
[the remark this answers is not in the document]
dapplion: Gone now

packages/ssz/src/viewDU/container.ts
dapplion: I don't if Javascript allows a better approach. All ideas I've found online are variations of this

packages/ssz/src/type/byteVector.ts
```diff
@@ -0,0 +1,84 @@
...
+ * - ViewDU: `Uint8Array`
+ *
+ * ByteVector is an immutable value which is represented by a Uint8Array for memory efficiency and performance.
+ * Note: Consumers of this type MUST never mutate the `Uint8Array` representation of a ByteVector.
+ */
+export class ByteVectorType extends ByteArrayType {
+  readonly typeName: string;
+  // Immutable characteristics
+  readonly depth: number;
+  readonly chunkDepth: number;
```
wemeetagain, line 24: should have a `chunkCount` property too

packages/ssz/src/type/union.ts
```diff
@@ -0,0 +1,198 @@
...
+    const minLens: number[] = [];
+    const maxLens: number[] = [];
+
+    for (const _type of types) {
+      minLens.push(_type.minSize);
+      maxLens.push(_type.maxSize);
+    }
+
+    this.minSize = 1 + Math.min(...minLens);
+    this.maxSize = 1 + Math.min(...maxLens);
```
wemeetagain, line 67: shouldn't this be Math.max ?

wemeetagain (changes requested):

packages/ssz/src/type/listComposite.ts
```diff
@@ -0,0 +1,172 @@
...
+    return tree_serializeToBytesArrayComposite(this.elementType, length, this.chunkDepth, chunksNode, output, offset);
+  }
+
+  tree_deserializeFromBytes(data: ByteViews, start: number, end: number): Node {
+    return tree_deserializeFromBytesArrayComposite(this.elementType, this.chunkDepth, data, start, end, this);
+  }
+
+  // Helpers for TreeView
+
+  tree_getLength(node: Node): number {
```
wemeetagain, line 129: is this method needed? same with listComposite

dapplion: Yes it's used by VIews

packages/ssz/src/type/union.ts
dapplion: Good pick! :+1:

packages/ssz/src/type/byteVector.ts
dapplion: Why? As far as I know it's not used by anyone

packages/persistent-merkle-tree/src/node.ts
```diff
@@ -126,3 +354,39 @@ export function compose(inner: Link, outer: Link): Link {
     return outer(inner(n));
   };
 }
+
+export function getNodeH(node: Node, hIndex: number): number {
+  if (hIndex === 0) return node.h0;
```
g11tech, line 359: will the switch here be faster? also not sure how hot this path is.

dapplion: Extremely hot, I did benchmarks with different options:
- if
- switch
- array to property

And if + switch had similar speeds. Makes sense because it is fundamentally the same operation. Both if and switch check every possible case before returning. Chose if because it looks more readable and concise.

dapplion:
I've squashed all commits into one for better visibility of this PR's comments and reviews. All commits will be squashed on merge so it's not important to preserve git history of this branch at this point

packages/ssz/src/type/composite.ts
```diff
@@ -0,0 +1,203 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  tree_createProof(node: Node, paths: Path[]): Proof {
+    throw Error("TODO");
+  }
+
+  protected abstract readonly maxChunkCount: number;
+  protected abstract getRoots(value: V): Uint8Array;
+}
+
+export abstract class TreeView<T extends CompositeType<unknown, unknown, unknown>> {
```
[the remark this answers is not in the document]
dapplion: Done

philknows:
Is this still getting merged in @dapplion as discussed on the Mar 18 planning call?

dapplion:
> Is this still getting merged in @dapplion as discussed on the Mar 18 planning call?

Yes @wemeetagain should help merge and release this

wemeetagain (approved):

PR #227, opened 2022-01-17
Add more benchmarks

dapplion:
**Motivation**

Add important benchmarks to test against https://github.com/ChainSafe/ssz/pull/223

**Description**

Add more benchmarks for:
- Full eth2 state (de)serialization
- More eth2 data structures


tuyennhv (approved):
lgtm, let's see how it works in ssz v2

PR #255, opened 2022-04-12
Add and use a new helper to digest64 two 32 bytes

g11tech:
**Motivation**
The ssz libraries when imported into browser by the light-client demo errors on the `Buffer.concat` functions.

This PR makes the library browser compatible for use by light-client demo.


packages/as-sha256/src/index.ts
```diff
@@ -39,6 +39,18 @@ export function digest64(data: Uint8Array): Uint8Array {
   throw new Error("InvalidLengthForDigest64");
 }
 
+export function digest64Bytes32(bytes1: Uint8Array,bytes2: Uint8Array): Uint8Array {
```
tuyennhv, line 42: maybe `digest64` = `digest2Bytes32`?

packages/as-sha256/src/index.ts
```diff
@@ -39,6 +39,18 @@ export function digest64(data: Uint8Array): Uint8Array {
   throw new Error("InvalidLengthForDigest64");
 }
 
+export function digest2Bytes32(bytes1: Uint8Array,bytes2: Uint8Array): Uint8Array {
```
dapplion, line 42: Inconsistent format, add space between arguments

dapplion (approved):

Issue #254, opened 2022-04-12
ssz `deserialize` modifies the input buffer


g11tech:
This was observed on de-serializing of a goerli-shadow-fork1 `bellatrix.SignedBeaconBlock`

```
x=fs.readFileSync("/data/unCompressedSnappy")
<Buffer 64 00 00 00 b6 9c 91 29 9d ad 7e 8a f1 1e 13 69 84 c4 42 4f 15 34 c3 31 39 53 a9 ea 17 9d 63 d9 01 7b 34 a5 5d ea e7 b1 b6 50 36 78 22 1d 99 86 ff 4c ... 139355 more bytes>
```
correctly deserializes x: 
```
> ltypes.ssz.bellatrix.SignedBeaconBlock.deserialize(x)
{
  message: {
    slot: 13249,
    proposerIndex: 18462,
```
second time call gives error because apparently x is modified: 
```
> ltypes.ssz.bellatrix.SignedBeaconBlock.deserialize(x)
Uncaught Error: Invalid deserialized bitlist, padding bit required
    at deserializeUint8ArrayBitListFromBytes (/usr/app/node_modules/@chainsafe/ssz/lib/type/bitList.js:96:15)
    at BitListType.deserializeUint8ArrayBitListFromBytes 
```
reload x  from file again, it works again.

also manually verified it, but comparing the buffer pre and post serialization, it gets modified at 8-10 places

Expected: The input buffer shouldn't be modified or if it speeds up the computation, a separate flag could be specified to work on  copy (true by default)

Workaround: A new buffer copy could be passed using `Buffer.from` on the original buffer.

PR #258, opened 2022-04-13
Force usage of Uint8Array.prototype.slice

dapplion:
**Motivation**

SSZ expected that all arguments typed as `Uint8Array` as `Uin8Array` but they could get a `Buffer` instance too.

Buffer.prototype.slice does not copy memory, force use Uint8Array.prototype.slice https://github.com/nodejs/node/issues/28087
 - Uint8Array.prototype.slice: Copy memory, safe to mutate
 - Buffer.prototype.slice: Does NOT copy memory, mutation affects both views

We could ensure that all Buffer instances are converted to Uint8Array before calling value_deserializeFromBytes. However doing that in a browser friendly way is not easy. Downstream code uses `Uint8Array.prototype.slice.call` to ensure Buffer.prototype.slice is never used. Unit tests also test non-mutability.

**Description**

Use `Uint8Array.prototype.slice` to ensure the method that copies memory is used. Using this alternative has no performance cost, since it's equivalent, also proof below:

```
      ✓ Copy Uint8Array 100000 slice                                        2863.475 ops/s    349.2260 us/op   x1.424       4897 runs   2.07 s
      ✓ Copy Uint8Array 100000 Uint8Array.prototype.slice.call              4144.614 ops/s    241.2770 us/op   x0.720      11186 runs   2.94 s
      ✓ Copy Buffer 100000 Uint8Array.prototype.slice.call                  4289.471 ops/s    233.1290 us/op   x1.053       7201 runs   2.00 s
```

I've added unit tests that ensure no mutability

Closes https://github.com/ChainSafe/ssz/issues/254

### DIFF
--- a/packages/ssz/src/type/abstract.ts
+++ b/packages/ssz/src/type/abstract.ts
@@ -124,6 +124,13 @@ export abstract class Type<V> {
 
   /** Deserialize binary data to value */
   deserialize(uint8Array: Uint8Array): V {
+    // Buffer.prototype.slice does not copy memory, force use Uint8Array.prototype.slice https://github.com/nodejs/node/issues/28087
+    // - Uint8Array.prototype.slice: Copy memory, safe to mutate
+    // - Buffer.prototype.slice: Does NOT copy memory, mutation affects both views
+    // We could ensure that all Buffer instances are converted to Uint8Array before calling value_deserializeFromBytes
+    // However doing that in a browser friendly way is not easy. Downstream code uses `Uint8Array.prototype.slice.call`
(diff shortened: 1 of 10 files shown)


dapplion:
If the `Uint8Array.prototype.slice` strategy doesn't work, we may go for .slice then .set on a new Uint8Array.

CC: @g11tech

g11tech (approved):

dapplion:
@g11tech and I have locally checked that the benchmark List[uint8, 68719476736] len 300000 ViewDU.get(i) does not degrade with this PR.

It's strange that 3 times in a row the benchmark is slower in CI by a factor of x6.

PR #263, opened 2022-04-26
Add EF multiproof format

wemeetagain:
**Motivation**

We need to support the multiproof format described in the consensus specs repo:
https://github.com/ethereum/consensus-specs/blob/dev/ssz/merkle-proofs.md

**Description**

Bare minimum to support EF multiproofs

TODO:
- [x] add tests

### DIFF
--- a/packages/persistent-merkle-tree/src/proof/index.ts
+++ b/packages/persistent-merkle-tree/src/proof/index.ts
@@ -1,5 +1,6 @@
 import {Gindex} from "../gindex";
 import {Node} from "../node";
+import {createEFMultiProof, createNodeFromEFMultiProof} from "./efMulti";
 import {createNodeFromSingleProof, createSingleProof} from "./single";
 import {
   computeTreeOffsetProofSerializedLength,
@@ -12,6 +13,7 @@ import {
 export enum ProofType {
   single = "single",
   treeOffset = "treeOffset",
+  EFMulti = "EFMulti",
 }
 
 /**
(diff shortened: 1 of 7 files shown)


dapplion:
Do we really have to support two multiproof formats? Can we can to consensus on which one is the canonical?

wemeetagain:
> Do we really have to support two multiproof formats?

Why not, its not something that changes. Once its implemented its done.

The two formats have different tradeoffs.

The tree offset format can be used for proofs which shape is not known ahead of time. Eg: give me the balance of the last validator
It has a 6ish% overhead in proof size tho, as the shape of the proof is usually serialized along with the proof (the `offsets` array). I haven't benchmarked proof generation / consumption speeds, but I suspect tree-offset proofs are faster to generate and consume. Would be interesting to see.

The EF format is "backwards compatible" with single branch proofs in the case of proving a single leaf.
It doesn't suffer the overhead, as the gindices of the proof is not normally serialized as part of the proof but rather agreed upon ahead of time.
This tradeoff makes it more suitable for proofs whose structure is known ahead of time, but makes it unsuitable for proofs whose shape is unknown by the consumer.

packages/persistent-merkle-tree/src/proof/index.ts
```diff
@@ -12,6 +13,7 @@ import {
 export enum ProofType {
   single = "single",
   treeOffset = "treeOffset",
+  EFMulti = "EFMulti",
```
dapplion, line 16: Then should we find a better name that's not tied to an organization? Some name that's generic to the data structure like "testOffset"

dapplion (approved):

PR #271, opened 2022-07-05
Implement single proof serialize/deserialize

hujw77:
**Motivation**

**Description**

Closes #issue_number


dapplion:
@hujw77 thank you so much for the contribution! Could you add tests to prove correct logic?

philknows:
We don't think we'll need proof serialization code in this library. Closing, thanks for the contribution though!

PR #278, opened 2022-09-03
Use allocUnsafe where possible

twoeths:
**Motivation**
- allocUnsafe() is better in NodeJS environment
- there is an external memory spike issue in lodestar

**Description**
- Use `allocUnsafe` where possible

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "@dapplion/benchmark": "^0.2.2",
     "@types/chai": "^4.2.15",
     "@types/mocha": "^8.2.2",
-    "@types/node": "^14.14.17",
+    "@types/node": "^16.11.55",
     "@typescript-eslint/eslint-plugin": "^4.19.0",
     "@typescript-eslint/parser": "^4.19.0",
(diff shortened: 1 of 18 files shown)


jeluard:
@tuyennhv this looks useful, wondering if that is still relevant?
Might be worth relying on [unint8arrays alloc](https://github.com/achingbrain/uint8arrays/blob/main/src/alloc.ts) for this. Also, we might want to consider their [pattern](https://github.com/achingbrain/uint8arrays/blob/main/src/xor.ts#L18) of converting back to `Uint8Array` to reduce `Buffer` inconsistencies.

philknows:
Bumping for update @twoeths

twoeths:
the idea was implemented for `as-sha256` in #391
for `ssz`, we gonna reuse memory a lot in the batch hash work so it's not necessary anymore
will work on this idea for `persistent-merkle-tree` and `lodestar`

twoeths:
`allocUnsafe()` in ssz is implemented in #409

PR #286, opened 2022-12-08
Allow to dynamically set type class name

dapplion:
**Motivation**

- See https://github.com/ChainSafe/lodestar/pull/4884

**Description**

Add `named()` static method with same arguments as constructor to dynamically set type class name


tuyennhv (approved):
looks good to me

wemeetagain (approved):
:100: looks great

PR #290, opened 2022-12-09
chore: yarn workspaces, release please

mpetrunic:
**Motivation**

**Description**

Closes #issue_number


wemeetagain:
Do we need to add a GITHUB_TOKEN? or will the one auto-injected be ok?

dapplion:
Why there's a 26,767 line diff in the yarn.lock?

mpetrunic:
> Why there's a 26,767 line diff in the yarn.lock?

Upgrade to new yarn version and format of the lockfile

mpetrunic:
> Do we need to add a GITHUB_TOKEN? or will the one auto-injected be ok?

Autoinjected is fine

dapplion (dismissed):

wemeetagain:
@ChainSafe/lodestar
After this is merged, you'll need to run `corepack enable` to properly use the correct version of `yarn` here.
See https://nodejs.org/api/corepack.html

wemeetagain (approved):

PR #292, opened 2022-12-16
feat: compact multiproof

wemeetagain:
**Motivation**

We want to support serving dynamic state/block multiproofs in Lodestar. The current approach is to request a proof as a list of "paths" into the state. Eg: `[ ["validators", 1000, "exitEpoch"], ["slot"] ]`. These paths are converted into generalized indices, which are then used as input in this library to create a proof.

This works for prototyping but it has several drawbacks.

First is that it is far from succinct as far as what's sent over the wire. Generally its just inefficient to send the textual representation of the requested items, even though it's convenient for human readability. There's also a lot of duplication when nested siblings are requested.

There's also a larger burden on the proof server, with more work to perform, and more that can go wrong. The paths must be validated (and in lodestar's case, converted to camel case). The paths must be converted into generalized indices. The helper indices must be calculated. All indices must be sorted for efficient traversal through the tree. And only then can the proof be generated.

In the longer term, we'd like a format that's more succinct and more performant for proof servers.

This is one such format. With this format, the request is in the form of a "proof descriptor", a bitstring, that describes the shape of the requested proof.
Not only is this very efficient over the wire, it also removes the vast majority of preprocessing work required by the proof server.

The proof descriptor only requires 2 bits per node included in the proof.
And the server only needs to validate the descriptor and can immediately generate the proof.

**Description**

Implementation of the proof format described in ethereum/consensus-specs#3148

See:
- The proof format: https://github.com/ethereum/consensus-specs/pull/3148
- Where it will be used: https://github.com/ethereum/beacon-APIs/pull/267

Results from the benchmarks:
```
    ✓ multiproof - depth 15, 1 requested leaves                           86445.37 ops/s    11.56800 us/op        -      60909 runs  0.809 s
    ✓ tree offset multiproof - depth 15, 1 requested leaves               38801.80 ops/s    25.77200 us/op        -      36924 runs   1.01 s
    ✓ compact multiproof - depth 15, 1 requested leaves                   205549.8 ops/s    4.865000 us/op        -      83819 runs   1.31 s

    ✓ multiproof - depth 15, 4 requested leaves                           23612.75 ops/s    42.35000 us/op        -      63635 runs   2.96 s
    ✓ tree offset multiproof - depth 15, 4 requested leaves               17984.96 ops/s    55.60200 us/op        -      18659 runs   1.11 s
    ✓ compact multiproof - depth 15, 4 requested leaves                   130293.2 ops/s    7.675000 us/op        -      19724 runs  0.808 s
```

### DIFF
--- a/packages/persistent-merkle-tree/src/proof/compactMulti.ts
+++ b/packages/persistent-merkle-tree/src/proof/compactMulti.ts
@@ -60,19 +60,19 @@ function getBit(bitlist: Uint8Array, bitIndex: number): boolean {
     case 0:
       return (byte & 0b1000_0000) !== 0;
     case 1:
-      return Boolean(byte & 0b0100_0000);
+      return (byte & 0b0100_0000) !== 0;
     case 2:
-      return Boolean(byte & 0b0010_0000);
+      return (byte & 0b0010_0000) !== 0;
     case 3:
-      return Boolean(byte & 0b0001_0000);
+      return (byte & 0b0001_0000) !== 0;
     case 4:
-      return Boolean(byte & 0b0000_1000);
+      return (byte & 0b0000_1000) !== 0;
     case 5:
-      return Boolean(byte & 0b0000_0100);
+      return (byte & 0b0000_0100) !== 0;
     case 6:
-      return Boolean(byte & 0b0000_0010);
+      return (byte & 0b0000_0010) !== 0;
     case 7:
-      return Boolean(byte & 0b0000_0001);
+      return (byte & 0b0000_0001) !== 0;
     default:
       throw new Error("unreachable");
   }
--- a/packages/persistent-merkle-tree/src/proof/compactMulti.ts
+++ b/packages/persistent-merkle-tree/src/proof/compactMulti.ts
@@ -58,7 +58,7 @@ function getBit(bitlist: Uint8Array, bitIndex: number): boolean {
   const byte = bitlist[byteIdx];
   switch (bit) {
     case 0:
-      return byte & 0b1000_0000 !== 0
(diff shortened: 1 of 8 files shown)


packages/persistent-merkle-tree/src/proof/compactMulti.ts
```diff
@@ -0,0 +1,151 @@
...
+  return descriptor;
+}
+
+function getBit(bitlist: Uint8Array, bitIndex: number): boolean {
+  const bit = bitIndex % 8;
+  const byteIdx = Math.floor(bitIndex / 8);
+  const byte = bitlist[byteIdx];
+  switch (bit) {
+    case 0:
+      return Boolean(byte & 0b1000_0000);
```
dapplion, line 61: Should be equivalent to
```suggestion
      return byte & 0b1000_0000 !== 0
```
? Type constructors are always a bit sus

dapplion (commented):
Overall looks good! Haven't asserted correctness against the spec. Minor comment

packages/persistent-merkle-tree/src/proof/compactMulti.ts
```diff
@@ -0,0 +1,151 @@
...
+  return descriptor;
+}
+
+function getBit(bitlist: Uint8Array, bitIndex: number): boolean {
+  const bit = bitIndex % 8;
+  const byteIdx = Math.floor(bitIndex / 8);
+  const byte = bitlist[byteIdx];
+  switch (bit) {
+    case 0:
+      return (byte & 0b1000_0000) !== 0;
```
dapplion, line 61: Should apply pattern consistently to all switch cases, one way or the other

dapplion (approved):

PR #296, opened 2023-01-04
feat: Switch case dependency to internal helper

acolytec3:
**Motivation**

Removes an external dependency, thus removing the surface area for supply chain atttacks.

**Description**

Replaces `case` dependency that's only used for string manipulation with an internal helper method that accomplishes the same goal.

Run impacted tests.

```sh
cd packages/ssz
npx mocha -- "test/unit/eth2/caseEth2.test.ts" "test/unit/strings.test.ts"
```

### DIFF
--- a/packages/ssz/src/util/strings.ts
+++ b/packages/ssz/src/util/strings.ts
@@ -1,0 +1,29 @@
+// Convert camelCase strings to various other formats  -- assumes input field is camelCase or camel1Case
+export const Case = {
+  snake: (field: string): string =>
+    field
+      .replaceAll(/[^0-z]/g, "")
+      .replaceAll(/[a-z][A-Z]|[0-9][A-Z]/g, (substr) => substr[0] + "_" + substr[1].toLowerCase()),
+  constant: (field: string): string =>
+    field
+      .replaceAll(/[^0-z]/g, "")
+      .replaceAll(/[a-z][A-Z]|[0-9][A-Z]/g, (substr) => substr[0] + "_" + substr[1])
+      .toUpperCase(),
+  pascal: (field: string): string => {
+    const first = field[0].toUpperCase();
+    return (first + field.slice(1)).replaceAll(/[^0-z]/g, "");
+  },
+  camel: (field: string): string => {
+    return field[0].toLowerCase() + field.slice(1);
+  },
+  header: (field: string): string => {
+    const first = field[0].toUpperCase();
+    return (
+      first +
+      field
+        .slice(1)
+        .replaceAll(/[^0-z]/g, "")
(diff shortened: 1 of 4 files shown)


dapplion (dismissed):
LGTM

packages/ssz/test/unit/strings.test.ts
```diff
@@ -0,0 +1,59 @@
+import {expect} from "chai";
+
+import {Case} from "../../src/util/strings";
+
+const testCases = ["fooBar", "foo1Bar"];
```
wemeetagain, line 5: Can you add a few more testcases? Maybe "fooBarBaz", "fooBar1Baz".

you can also make tests more generic like:
```ts
const testCases = [
  { input: "fooBar", snake: "foo_bar", pascal: "FooBar" ......},
  ...
]
const cases = ["snake", "pascal", .....]

for (const case of cases) {
  describe(`should convert strings to ${case} case`, () => {
    for (const testCase of testCases) {
      const expected = testCase[case];
      it(`should convert to ${expected}`, () => {
        const field = Case[case](testCase.input);
        expect(field).to.equal(expected);
      });
    }
  })
}
```

wemeetagain (changes requested):

g11tech (commented):
looks good :+1:

packages/ssz/test/unit/strings.test.ts
acolytec3: Done.  I think I got all the typing correct too but let me know if you want a different approach on this.

packages/ssz/src/util/strings.ts
```diff
@@ -0,0 +1,31 @@
...
+    const first = field[0].toUpperCase();
+    return (
+      first +
+      field
+        .slice(1)
+        .replaceAll(/[^0-z]/g, "")
+        .replaceAll(/[a-z][A-Z]|[0-9][A-Z]/g, (substr) => substr[0] + "-" + substr[1])
+    );
+  },
+  eth2: (field: string): string => Case.snake(field),
```
wemeetagain, line 30: Can you incorporate https://github.com/ChainSafe/ssz/pull/296/files#diff-f75d38629191b6e32405d459655ce488f3d05f36e05e548a87fb3306f80ef72dR553-R555 here? so that can be simplified

acolytec3: I think your deeplink might be broken but are you referring to [this line](https://github.com/ChainSafe/ssz/blob/b783bfbea4f456497950d8724937b098873cf2ef/packages/ssz/src/type/container.ts#L554)?  If so, just folded it in with my latest commit.

wemeetagain (dismissed):

wemeetagain (approved):

g11tech (approved):

Issue #313, opened 2023-03-22
Replace wasm as-sha256 with pure js noble-hashes


paulmillr:
1. wasm is not supported in all environments. the library is used in ethereumjs, which means the use-cases can be extensive
2. as-sha256 is slower than pure-js audited implementation in https://github.com/paulmillr/noble-hashes. Specifically, on inputs bigger than 64 bytes, Noble always wins. JS is 43% faster than as-sha256. Proof:

```
SHA256 32B @chainsafe/as-sha256 x 1,615,508 ops/sec @ 619ns/op ± 3.57% (min: 500ns, max: 2ms)
SHA256 32B noble x 1,175,088 ops/sec @ 851ns/op ± 1.42% (min: 667ns, max: 962μs)

SHA256 64B @chainsafe/as-sha256 x 1,257,861 ops/sec @ 795ns/op ± 2.80% (min: 666ns, max: 2ms)
SHA256 64B noble x 822,368 ops/sec @ 1μs/op ± 1.73% (min: 1μs, max: 742μs)

SHA256 1KB @chainsafe/as-sha256 x 122,234 ops/sec @ 8μs/op ± 3.61% (min: 7μs, max: 7ms)
SHA256 1KB noble x 165,837 ops/sec @ 6μs/op

SHA256 8KB @chainsafe/as-sha256 x 16,497 ops/sec @ 60μs/op ± 4.12% (min: 56μs, max: 7ms)
SHA256 8KB noble x 23,789 ops/sec @ 42μs/op

SHA256 1MB @chainsafe/as-sha256 x 132 ops/sec @ 7ms/op
SHA256 1MB noble x 187 ops/sec @ 5ms/op
```

(it's slightly slower for 32B inputs, but executing 1 million hashes of 32b items at once is not a common case)

3. wasm security is not as good as it can be thought of

dapplion:
I see your concerns and would love to find a setup to benefits both web and nodejs consumers. For Lodestar's beacon node the speed of 64B hashes is crucial for performance since it's one of the most frequent ops we do.

@FrederikBolding would there be some approach like platform aware entrypoints that would fix your build issue? Say a-la https://www.npmjs.com/package/cross-fetch

FrederikBolding:
[...]

PR #314, opened 2023-03-25
feat: add swappable hasher, default to noble-hashes

wemeetagain:
**Motivation**

See #313 

We can replace as-sha256 with noble-hashes as long as we support our current interfaces and expose some mechanism to update the implementation for Lodestar's sake.

Its likely overkill to pass hash functions down thru our stack, and we can assume that a runtime will only require a single hash implementation.

**Description**

- Replace as-sha256 (and avoid loading any wasm) with noble-hashes sha256 implementation by default.
- Add a swapable "hasher" that can be updated before importing the persistent-merkle-tree / ssz or consumer libraries

Swapping the hasher with as-sha256, as will be required by lodestar, can be done like so:
```ts
// in lodestar, before anything else

import {hasher} from "@chainsafe/persistent-merkle-tree/hasher/as-sha256";
import {setHasher} from "@chainsafe/persistent-merkle-tree/hasher";

setHasher(hasher);

await import("./the-rest-of-the-application.js");
```

**Benchmarks**

Feel free to figure out how to speed this up if desired. I might be doing something stupid here.

```
yarn benchmark:files packages/persistent-merkle-tree/test/perf/hasher.test.ts

  hasher
    ✓ hash 2 Uint8Array 2250026 times - as-sha256                        0.4682371 ops/s    2.135670  s/op        -         12 runs   27.7 s
    ✓ hashTwoObjects 2250026 times - as-sha256                           0.4947299 ops/s    2.021305  s/op        -         12 runs   26.3 s
    ✓ hash 2 Uint8Array 2250026 times - noble                            0.1935327 ops/s    5.167086  s/op        -         11 runs   61.7 s
    ✓ hashTwoObjects 2250026 times - noble                               0.1418410 ops/s    7.050149  s/op        -          8 runs   63.4 s
```


packages/persistent-merkle-tree/src/hasher/noble.ts
```diff
@@ -0,0 +1,10 @@
+import {sha256} from "@noble/hashes/sha256";
+import type {Hasher} from "./types";
+import {hashObjectToUint8Array, uint8ArrayToHashObject} from "./util";
+
+const digest64 = (a: Uint8Array, b: Uint8Array): Uint8Array => sha256.create().update(a).update(b).digest();
```
paulmillr, line 5: see if this helps:

```
const digest64 = (a: Uint8Array, b: Uint8Array): Uint8Array => sha256(concatBytes(a, b))
```

concatBytes can be imported from noble/hashes/utils

g11tech: would it give perf boost, from my understanding the underlying lib (for e.g. as-sha256) will be calling chunkified updates itself if you pass bigger input

g11tech (commented):
looks amazing to me, leaving it unapproved if there is a followup on the perf improvement

.eslintrc.js
```diff
@@ -81,6 +81,9 @@ module.exports = {
 
     // Prevents accidentally pushing a commit with .only in Mocha tests
     "no-only-tests/no-only-tests": "error",
+
+    // TEMP Disabled while eslint-plugin-import support ESM (Typescript does support it) https://github.com/import-js/eslint-plugin-import/issues/2170
```
mpetrunic, line 85: Have you tried https://github.com/import-js/eslint-import-resolver-typescript ?

packages/persistent-merkle-tree/src/hasher/noble.ts
paulmillr: probably won't, but just in case

paulmillr:
I wonder why the perf difference here is 5x, while noble-hashes benchmarks only show 1.5x on 64b. Are you confident the benchmarks are proper?

wemeetagain:
> I wonder why the perf difference here is 5x, while noble-hashes benchmarks only show 1.5x on 64b. Are you confident the benchmarks are proper?

We mostly hash two HashObjects, which is 3.5x. Not quite to 5x but in the ballpark.

dapplion:
@wemeetagain can you set the benchmarks to use the WASM version? Else we go blind for Lodestar performance tracking

dapplion:
Thanks @wemeetagain benchmarks look like before now :+1:

paulmillr:
Could you merge this please? Current wasm breaks unsupported environments.

wemeetagain:
> Could you merge this please?

Yes, looking for a :heavy_check_mark: from @ChainSafe/lodestar and will cut a release afterwards

dapplion (approved):

Issue #322, opened 2023-05-02
Show errors during serialization/deserialization


nflaig:
@g11tech yes was able to verify it works and looks correct locally, something I noticed during testing is that there is no error shown when trying to serialize/deserialize invalid data, it only logs an error to the console. This is also the case on the live website (https://simpleserialize.com/).

For example:

```
caught (in promise) RangeError: Offset is outside the bounds of the DataView
    at DataView.getUint32 (<anonymous>)
    at readVariableOffsets (container.js:322:1)
    at ContainerType.getFieldRanges (container.js:290:1)
    at ContainerType.value_deserializeFromBytes (container.js:125:1)
    at ContainerType.deserialize (abstract.js:45:1)
    at Object.parse (input_types.ts:21:32)
    at Input._callee3$ (Input.tsx:117:30)
    at tryCatch (runtime.js:63:16)
    at Generator.invoke [as _invoke] (runtime.js:294:1)
    at Generator.next (runtime.js:119:1)
```

Maybe catch the error and show a simple browser alert?

_Originally posted by @nflaig in https://github.com/ChainSafe/ssz/issues/321#issuecomment-1531639631_

PR #323, opened 2023-05-02
fix: show errors during serialization/deserialization as popup

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/ssz/issues/322

**Description**

Easiest solution I found with high confidence of not breaking something.

It will now also show a error popup instead of just logging to console, the errors are quite technical and might not be that helpful sometimes but at least a user gets notified that something is not right.

image


g11tech (approved):
lgtm

Issue #324, opened 2023-05-02
Running `yarn build` from root directory fails on master branch


nflaig:
I just ran `yarn build` from root directory, looks like there is a build issue, I haven't faced that before because I only ran build in sub directory for website specifically

```
➤ YN0000: [simpleserialize.com]: ERROR in ../ssz/lib/util/merkleize.js 4:16-77
➤ YN0000: [simpleserialize.com]: Module not found: Error: Package path ./lib/hasher/index is not exported from package /home/nico/projects/ethereum/ssz/packages/ssz/node_modules/@chainsafe/persistent-merkle-tree (see exports field in /home/nico/projects/ethereum/ssz/packages/ssz/node_modules/@chainsafe/persistent-merkle-tree/package.json)
➤ YN0000: [simpleserialize.com]:  @ ../ssz/lib/index.js 63:18-45
➤ YN0000: [simpleserialize.com]:  @ ./src/components/Input.tsx 75:0-45 178:26-37 462:24-35
➤ YN0000: [simpleserialize.com]:  @ ./src/components/Serialize.tsx 46:0-28 219:42-47
➤ YN0000: [simpleserialize.com]:  @ ./src/components/Tabs.tsx 37:0-36 88:105-114
➤ YN0000: [simpleserialize.com]:  @ ./src/App.tsx 4:0-37 7:193-197
➤ YN0000: [simpleserialize.com]:  @ ./src/index.tsx 7:0-24 23:46-49
➤ YN0000: [simpleserialize.com]: 
➤ YN0000: [simpleserialize.com]: ERROR in ../ssz/lib/util/zeros.js 4:16-77
➤ YN0000: [simpleserialize.com]: Module not found: Error: Package path ./lib/hasher/index is not exported from package /home/nico/projects/ethereum/ssz/packages/ssz/node_modules/@chainsafe/persistent-merkle-tree (see exports field in /home/nico/projects/ethereum/ssz/packages/ssz/node_modules/@chainsafe/persistent-merkle-tree/package.json)
➤ YN0000: [simpleserialize.com]:  @ ../ssz/lib/util/merkleize.js 5:16-34
➤ YN0000: [simpleserialize.com]:  @ ../ssz/lib/index.js 63:18-45
➤ YN0000: [simpleserialize.com]:  @ ./src/components/Input.tsx 75:0-45 178:26-37 462:24-35
[...]

PR #318, opened 2023-04-19
fix: Use file path imports

FrederikBolding:
**Motivation**

The goal of this PR is to stop relying on `package.json` `exports`, as it may break older build tools such as Browserify.

**Description**

Uses explicit imports for all imports that were previously using `exports` aliases.

**Steps to test or reproduce**

PR #325, opened 2023-05-03
fix: add hasher alias to webpack config to fix build

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/ssz/issues/324

**Description**

Add alias to webpack config to fix build.

@g11tech I noticed you [added this](https://github.com/ChainSafe/ssz/compare/a7548d9b9fc5ed4315eaca8aebb8db4970e48c7e..c57882e2383e17399bed01e02512631d2f13e55a) in https://github.com/ChainSafe/ssz/pull/318 but it got removed by  a force-push.

### DIFF
--- a/packages/simpleserialize.com/webpack.config.js
+++ b/packages/simpleserialize.com/webpack.config.js
@@ -17,6 +17,10 @@ const config = {
   },
   resolve: {
     extensions: ['.js', '.jsx', '.ts', '.tsx'],
+    alias: {
+      '@chainsafe/persistent-merkle-tree/lib/hasher/index':
+        '../../node_modules/@chainsafe/persistent-merkle-tree/lib/hasher/noble',
+    },
   },
   module: {
     rules: [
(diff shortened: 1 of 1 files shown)


nflaig:
looks like this fixes it locally for me but breaks it in CI, must be missing something else

g11tech (approved):
lgtm, although aliasing shouldn't be required anymore (or that was the aim)

nflaig:
@g11tech yeah, looking at https://github.com/ethereumjs/ethereumjs-monorepo/pull/2656, seems to be working fine by removing aliases. The build only fails for me locally, CI works as expected. Are you able to build locally? I must be missing something

g11tech:
i have figured out what is happening, old cached node modules of previous versions are still interfering 

1. Remove all node_modules ( main repo root `node_modules` as well as  inside `packages/*/node_modules` If you had old code, their previous builds are cached there and are interfering)
2. `corepack enable` if you haven't been running it to enable yarn workspace functionality

then 
1. `yarn --immutable` for installing packages
2. `yarn build`

then the build is successful

Issue #326, opened 2023-07-07
Support full Uint64 in simpleserialize.com


wemeetagain:
**Is your feature request related to a problem? Please describe.**

See https://discord.com/channels/595666850260713488/598292067260825641/1126886479898099752
"I found that tool very useful to check against something working when trying to find root mismatches
but most static tests would fall out of the 53 bit limits"

**Describe the solution you'd like**

Support the full uint64 range on the website

PR #327, opened 2023-07-07
feat: support full uint64 range on simpleserialize.com

nflaig:
**Motivation**

Closes https://github.com/ChainSafe/ssz/issues/326

**Description**

Patch SSZ types to support full uint64 range on simpleserialize.com


wemeetagain (approved):
LGTM, this is the same approach we use to pass the ssz_static spec tests

PR #328, opened 2023-07-07
fix: return correct property gindex for ContainerNodeStructType

wemeetagain:
**Motivation**

While doing gindex calculations on a BeaconState, noticed that there is a 'bug' in getting gindexes for validator entries.

**Description**

Remove the overridden getPropertyGindex that prevents calculation of gindices on a ContainerNodeStructType

### DIFF
--- a/packages/ssz/src/type/containerNodeStruct.ts
+++ b/packages/ssz/src/type/containerNodeStruct.ts
@@ -77,9 +77,6 @@ export class ContainerNodeStructType<Fields extends Record<string, Type<unknown>
 
   // ContainerNodeStructType can only parse proofs that contain all the data.
   // TODO: Support converting a partial tree to a partial value
-  getPropertyGindex(): null {
-    return null;
-  }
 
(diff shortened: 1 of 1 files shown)


packages/ssz/src/type/containerNodeStruct.ts
```diff
@@ -77,9 +77,6 @@ export class ContainerNodeStructType<Fields extends Record<string, Type<unknown>
 
   // ContainerNodeStructType can only parse proofs that contain all the data.
   // TODO: Support converting a partial tree to a partial value
-  getPropertyGindex(): null {
-    return null;
-  }
```
dapplion, line 82: It's been a while since I dealt with this code, but what about the comment?

> ContainerNodeStructType can only parse proofs that contain all the data

wemeetagain: I agree with the comment. I can put in a more effort to make the UX a bit nicer here

PR #329, opened 2023-08-27
feat: add optional ssz type

g11tech:
implement PR:
 - [eip-6475](https://eips.ethereum.org/EIPS/eip-6475)

requirement to build verkle types as the kaustenine network (and its relaunch will use optional types)

### DIFF
--- a/packages/ssz/src/type/optional.ts
+++ b/packages/ssz/src/type/optional.ts
@@ -3,7 +3,7 @@ import {mixInLength} from "../util/merkleize";
 import {Require} from "../util/types";
 import {namedClass} from "../util/named";
 import {Type, ByteViews, JsonPath, JsonPathProp} from "./abstract";
-import {CompositeType, CompositeTypeAny, isCompositeType} from "./composite";
+import {CompositeType, isCompositeType} from "./composite";
 import {addLengthNode, getLengthFromRootNode} from "./arrayBasic";
 /* eslint-disable @typescript-eslint/member-ordering */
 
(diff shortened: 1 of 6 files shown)


packages/ssz/src/type/optional.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+  }
+
+  getPropertyType(): Type<unknown> {
+    return this.elementType;
+  }
+
+  getIndexProperty(index: number): string | number | null {
+    if (isCompositeType(this.elementType)) {
+      return this.elementType.getIndexProperty(index);
```
g11tech, line 169: may be an de-concatenation op is required (to remove `VALUE_GINDEX` before passing in the index to the underlying element type

wemeetagain: I think this method should only return `"value"` or `"selector"` for 0 or 1?

g11tech: transparent because `value` and `selector` properties don't exist as per my understanding

wemeetagain:
@g11tech I did a review of this PR and added some commits on top, let me know what you think.

wemeetagain (approved):

packages/ssz/src/type/optional.ts
```diff
@@ -55,24 +59,34 @@ export class OptionalType<ElementType extends Type<unknown>> extends CompositeTy
     return null as ValueOfType<ElementType>;
   }
 
-  getView(tree: Tree): ValueOfType<ElementType> {
-    return this.tree_toValue(tree.rootNode);
+  // TODO fix
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  getView(_tree: Tree): ValueOfType<ElementType> {
+    throw new Error("not implemented");
```
g11tech: @wemeetagain should we not depending on its composite type or not provide the toValue or underlying view?

g11tech:
have merged for now @wemeetagain , we can try adding other view/tree methods if there is need :+1: 

Also as per the plan, the executionWitness might get serialized later on for embedding in executionPayload so we might not need this at all  (but i guess ethereumjs and/or other EL side codebases would be able to use this, so good for ecosystem in general)

PR #333, opened 2023-09-13
chore: copy latest ssz types from lodestar

tuyennhv:
**Motivation**

Test latest lodestar types within ssz

**Description**

- Copy latest ssz lodestar types
- Use latest spec test in lodestar


g11tech (approved):
lgtm

PR #334, opened 2023-09-14
feat: publish more methods

tuyennhv:
**Motivation**

- We want `getFieldRanges` to be published so that we can load different chunks of data on demand in lodestar
- We also want to use `deserializeUint8ArrayBitListFromBytes` in lodestar https://github.com/ChainSafe/lodestar/blob/v1.11.1/packages/beacon-node/src/util/sszBytes.ts#L220

**Description**

Publish the above methods in ssz


dapplion (approved):

matthewkeil (approved):
LGTM 🚀

wemeetagain (approved):

PR #336, opened 2023-09-18
feat: implement sliceTo() for ListBasicTreeViewDU

tuyennhv:
**Motivation**

- For lodestar, we may want to load beacon state ssz bytes to an existing `ViewDU` object, this is needed to optimize the process in case a field is rarely changed

**Description**

- Add `sliceTo` method to ListBasicTreeViewDU, similar to https://github.com/ChainSafe/ssz/blob/d18ea08177544c82b9a1e7dd09127b4d63a2b2d8/packages/ssz/src/viewDU/listComposite.ts#L45


g11tech (approved):
lgtm

dapplion (changes requested):
Implementation logic looks correct and good. I would only drop support for negative integers

dapplion (approved):

PR #347, opened 2024-02-28
fix: improve digest64 for as-sha256

tuyennhv:
**Motivation**

Improve `digest64`

**Description**

- Use latest `assemblyscript` yields better performance
- Chain `w` computation to the main loop in `hashBlocks()`
- Return result by using `Uint8Array.slice()`

Before:
```
digestTwoHashObjects vs digest64 vs digest
    ✓ digestTwoHashObjects 50023 times                                    27.71683 ops/s    36.07916 ms/op   x1.042        264 runs   10.1 s
    ✓ digest64 50023 times                                                26.85761 ops/s    37.23339 ms/op   x1.020        256 runs   10.1 s
    ✓ digest 50023 times                                                  27.02430 ops/s    37.00373 ms/op   x1.010        259 runs   10.1 s

```

After
```
digestTwoHashObjects vs digest64 vs digest
    ✓ digestTwoHashObjects 50023 times                                    30.54306 ops/s    32.74066 ms/op   x0.946        293 runs   10.1 s
    ✓ digest64 50023 times                                                28.28273 ops/s    35.35727 ms/op   x0.968        271 runs   10.1 s
    ✓ digest 50023 times                                                  27.93469 ops/s    35.79778 ms/op   x0.977        268 runs   10.1 s
  
```

`digestTwoHashObjects` is ~10% faster and `digest64` is ~5% faster


tuyennhv:
the benchmark result in CI is not precise because it run with very short period of time, updated it to run at least 10s

```typescript
setBenchOpts({
    minMs: 10000,
  });
```

packages/as-sha256/package.json
```diff
@@ -30,26 +30,21 @@
...
+    "test": "yarn run test:unit",
     "test:unit": "yarn run test:unit:node && yarn run test:unit:browser",
     "test:unit:node": "run -T mocha -r ts-node/register test/unit/*.test.ts",
     "test:unit:browser": "run -T karma start --single-run --browsers ChromeHeadless,FirefoxHeadless karma.config.js",
     "benchmark": "node -r ts-node/register ./node_modules/.bin/benchmark 'test/perf/index.test.ts'",
     "benchmark:local": "yarn benchmark --local",
-    "test:ci": "yarn test:as-ci",
-    "test:as": "asp --nortrace --verbose",
-    "test:as-ci": "asp --nortrace 2> /dev/null"
+    "test:ci": "yarn test:as-ci"
```
jeluard, line 43: `test:as-ci` has been removed?

packages/as-sha256/test/perf/index.test.ts
```diff
@@ -1,12 +1,17 @@
-import {itBench} from "@dapplion/benchmark";
+import {itBench, setBenchOpts} from "@dapplion/benchmark";
 import * as sha256 from "../../src";
 
-// Aug 10 2021
+// Feb 2024 Mac M1
 // digestTwoHashObjects vs digest64 vs digest
-//     ✓ digestTwoHashObjects 50023 times                                    19.17366 ops/s    52.15488 ms/op        -       1151 runs   60.0 s
```
jeluard, line 6: Side-remark: previous benchmark were run for `60S`, maybe we want to align with that.

packages/as-sha256/assembly/index.ts
```diff
@@ -141,6 +141,7 @@ function SIG1(x: u32): u32 {
   return rotr(x, 17) ^ rotr(x, 19) ^ (x >>> 10);
 }
 
+var tmpW: u32;
```
jeluard, line 144: Nitpick: prefer `let` ?

packages/as-sha256/assembly/index.ts
```diff
@@ -158,25 +159,17 @@ function hashBlocks(wPtr: usize, mPtr: usize): void {
...
-  }
+  // 16 first u32 of expanded message block are same as message block
+  // rest of the 48 u32 are computed from the first 16 u32
+  // instead of computing expanded messsage blocks in a separate for loop,
+  // chain it to the below for loop to improve performance
 
   // Apply SHA256 compression function on expanded message blocks
   for (i = 0; i < 64; i++) {
-    t1 = h + EP1(e) + CH(e, f, g) + load32(kPtr, i) + load32(wPtr, i);
+    tmpW = i < 16 ? load32be(mPtr, i) : SIG1(load32(wPtr, i - 2)) + load32(wPtr, i - 7) +
```
jeluard, line 169: It looks like `tmpW` could be declared inside the loop? Like `const tmpW = ..`. It appears to be only assigned once per iteration and not updated.

jeluard: Or is it some assemblyscript specifics?

jeluard (commented):
Do we need to store `build` artifacts in the repo?

packages/as-sha256/package.json
tuyennhv: there is no `as` tests for now and that causes tests not to be scanned by CI for a while due to this. Removing it helps run tests again in CI

packages/as-sha256/assembly/index.ts
tuyennhv: we don't want to allocate variables there to avoid `gc`, it's designed like that for all other places as well.
yes, it's specific to assemblyscript

packages/as-sha256/assembly/index.ts
tuyennhv: yeah good catch, will apply for other places as well

tuyennhv:
> Do we need to store `build` artifacts in the repo?

yes, they are actually input for our typescript function every time we run, that's why they are tracked in git

packages/as-sha256/package.json
jeluard: Wouldn't calling `yarn test:as-ci` fail then? As `test:ci` is calling it?
`test:ci` doesn't appear to be used (only `test`) so maybe it could be removed too?

packages/as-sha256/package.json
```diff
@@ -30,26 +30,21 @@
...
   },
   "devDependencies": {
-    "@as-pect/assembly": "2.8.1",
```
jeluard, line 48: Can the file `as-pect.config.js` be removed too?

jeluard:
> > Do we need to store `build` artifacts in the repo?
> 
> yes, they are actually input for our typescript function every time we run, that's why they are tracked in git

Is [this](https://github.com/ChainSafe/ssz/blob/master/packages/as-sha256/scripts/codegen.ts#L7) what you are referring to? Looks like we only care about `codegen` being called to update https://github.com/ChainSafe/ssz/blob/master/packages/as-sha256/src/wasmCode.ts (as correctly done by this PR), but wasm/wat don't need to be in git?

Not part of this PR anyway, maybe something to be considered to improve the build process.

tuyennhv:
checked the performance on a separate server and got same result to my local environment

master feat4 mainnet

```
digestTwoHashObjects vs digest64 vs digest
    ✓ digestTwoHashObjects 50023 times                                    19.25844 ops/s    51.92528 ms/op        -         67 runs   3.98 s
    ✓ digest64 50023 times                                                20.23583 ops/s    49.41730 ms/op        -         12 runs   1.11 s
    ✓ digest 50023 times                                                  20.31116 ops/s    49.23402 ms/op        -         13 runs   1.19 s
```
this branch

```
digestTwoHashObjects vs digest64 vs digest
    ✓ digestTwoHashObjects 50023 times                                    21.91978 ops/s    45.62089 ms/op        -       1304 runs   60.0 s
    ✓ digest64 50023 times                                                21.11148 ops/s    47.36759 ms/op        -       1256 runs   60.0 s
    ✓ digest 50023 times                                                  20.90464 ops/s    47.83628 ms/op        -       1244 runs   60.0 s
```

however benchmark result in CI consistently shows worse statistics so need to investigate this, may need to separate to smaller PRs in order to figure out the issue

PR #349, opened 2024-03-01
feat: add capability to cache merkle roots for lists

g11tech:
need to cache merkle root of big lists (Transactions), as well as to allow setting the cached root (so needing cached symbol import) so that if we have for e.g. transactionsRoot, that can be set on transactions helping avoid merklelization of it in big blocks


tuyennhv (approved):
looks good to me, was thinking we can make a util to be used for `compisite.ts`, `listComposite.ts` and `listBasic.ts` but seems like they have different type

PR #350, opened 2024-03-06
feat: improve ViewDU.serialize()

tuyennhv:
**Motivation**

- ViewDU has cached nodes, however when we serialize we go through the respective type and it loads all nodes again, see https://github.com/ChainSafe/ssz/blob/6220d320b004ea80bd30925487ac0f3299295528/packages/ssz/src/type/container.ts#L233
- This is to improve `state.serialize()` in lodestar: right after we get through state transition, we always have cached nodes inside

**Description**
Define `serialize()` in ViewDU to go through the new `serializeToBytes()` method instead of its respective type class

- Add `serializeToBytes()` method to TreeViewDU, by default it'll delegate to respective type class
- The container ViewDU `serializeToBytes()` will recursively call respective `serializeToBytes` of internal ViewDU properties
- For `listComposite` and `listBasic` ViewDUs, they'll use use cached nodes if possible
- Also export some util types so that lodestar can consume

**Test**
New `serialize()` implementation in ViewDU is scanned through spec tests

### DIFF
--- a/packages/ssz/src/viewDU/container.ts
+++ b/packages/ssz/src/viewDU/container.ts
@@ -1,5 +1,5 @@
 import {getNodeAtDepth, LeafNode, Node, setNodesAtDepth} from "@chainsafe/persistent-merkle-tree";
-import {Type} from "../type/abstract";
+import {ByteViews, Type} from "../type/abstract";
 import {BasicType, isBasicType} from "../type/basic";
 import {CompositeType, isCompositeType, CompositeTypeAny} from "../type/composite";
 import {ContainerTypeGeneric} from "../view/container";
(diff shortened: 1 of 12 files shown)


packages/ssz/src/viewDU/container.ts
```diff
@@ -114,6 +114,44 @@ class ContainerTreeViewDU<Fields extends Record<string, Type<unknown>>> extends
...
+        node = getNodeAtDepth(this._rootNode, this.type.depth, index);
+        this.nodes[index] = node;
+      }
+      if (fieldType.fixedSize === null) {
+        // write offset
+        output.dataView.setUint32(fixedIndex, variableIndex - offset, true);
+        fixedIndex += 4;
+        // write serialized element to variable section
+        // basic types always have fixedSize
+        if (isCompositeType(fieldType)) {
```
g11tech, line 139: do we have some test case(s) covering all the if conditions here?

tuyennhv: yeah the existing test cases implicitly covered this but it's good to have an explicit test case for it, I added it. `runViewTestMutation` compares `type.serialize()` vs `ViewDU.serialize()`
also spec test with `eth2` types scanned through the code too

g11tech (approved):
lgtm

tuyennhv:
Implementation note for lodestar: use `BeaconStateViewDU.serialize()`, or the `ViewDU.serializeToBytes()` if there is a preallocated buffer

PR #352, opened 2024-03-08
feat: new type for list of uint64

tuyennhv:
**Motivation**

- Improve performance of `ListBasicType<UintNumberType<8>>.toViewDU()` (`balances` type in lodestar), the performance issue was noted here https://github.com/ChainSafe/ssz/blob/14c4457026a9fbea5bfe5c66580f7c8a8bee790a/packages/ssz/src/type/abstract.ts#L96
- Cache the created leaf nodes when calling that function and populate them to resulting ViewDU object, see https://github.com/ChainSafe/lodestar/issues/6520

**Description**

- New `ListUintNum64Type` class and overwrite `toViewDU()` function:
  - create nodes directly from `number[]` instead of having to `serialize()/deserialize()` implementation in the parent abstract class
  - use the created LeftNodes to populate the final ViewDU object

part of https://github.com/ChainSafe/lodestar/issues/6520

### DIFF
--- a/packages/persistent-merkle-tree/src/packedNode.ts
+++ b/packages/persistent-merkle-tree/src/packedNode.ts
@@ -1,9 +1,43 @@
 import {subtreeFillToContents} from "./subtree";
 import {Node, LeafNode, getNodeH, setNodeH} from "./node";
 
+const NUMBER_2_POW_32 = 2 ** 32;
+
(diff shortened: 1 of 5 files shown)


wemeetagain (approved):
LGTM

tuyennhv:
Implementation note for lodestar: migrate `state.balances` to `export const Balances = new ListUintNum64Type(VALIDATOR_REGISTRY_LIMIT);`

PR #366, opened 2024-04-03
feat: add `sliceFrom` to `ListCompositeTreeViewDU`

ensi321:
Add `sliceFrom` to `ListCompositeTreeViewDU`. 

Used in EIP-7251 to slice various beacon state fields like `pendingBalanceDeposit`, `pendingConsolidation` during epoch processing. 

Refer to the spec for more details:
[`process_pending_balance_deposits`](https://github.com/ethereum/consensus-specs/blob/f5207db2c415f676bb61d173031d1c233579f1a8/specs/_features/eip7251/beacon-chain.md?plain=1#L599)
[`process_pending_consolidations `](https://github.com/ethereum/consensus-specs/blob/f5207db2c415f676bb61d173031d1c233579f1a8/specs/_features/eip7251/beacon-chain.md?plain=1#L628)
[`process_withdrawals`](https://github.com/ethereum/consensus-specs/blob/f5207db2c415f676bb61d173031d1c233579f1a8/specs/_features/eip7251/beacon-chain.md?plain=1#L734)

### DIFF
--- a/packages/ssz/src/viewDU/listComposite.ts
+++ b/packages/ssz/src/viewDU/listComposite.ts
@@ -89,7 +89,11 @@ export class ListCompositeTreeViewDU<
     let newChunksNode;
     let newLength;
 
-    if (index >= this.nodes.length) {
+    if (index > this.nodes.length) {
+      throw new Error(`Does not support sliceFrom() with index out of bound ${index}`);
+    }
+
+    if (index === this.nodes.length) {
       newChunksNode = zeroNode(this.type.chunkDepth);
       newLength = 0;
     } else {
--- a/packages/ssz/src/viewDU/listComposite.ts
+++ b/packages/ssz/src/viewDU/listComposite.ts
@@ -71,7 +71,7 @@ export class ListCompositeTreeViewDU<
    *
    * ```ts
    * const nodes = getChunkNodes()
-   * return listFromChunkNodes(node.clice(index))
+   * return listFromChunkNodes(node.slice(index))
    * ```
    *
    * Note: If index === n, returns an empty list of length 0
--- a/packages/ssz/src/viewDU/listComposite.ts
+++ b/packages/ssz/src/viewDU/listComposite.ts
@@ -67,8 +67,7 @@ export class ListCompositeTreeViewDU<
   }
 
   /**
-   * Returns a new ListCompositeTreeViewDU instance with the values from `index` to `n`
-   * where n is the length of the list.
+   * Returns a new ListCompositeTreeViewDU instance with the values from `index` to the end of list
    * 
    * ```ts
    * const nodes = getChunkNodes()
--- a/packages/ssz/src/viewDU/listComposite.ts
+++ b/packages/ssz/src/viewDU/listComposite.ts
@@ -68,12 +68,12 @@ export class ListCompositeTreeViewDU<
 
   /**
    * Returns a new ListCompositeTreeViewDU instance with the values from `index` to the end of list
-   * 
+   *
    * ```ts
    * const nodes = getChunkNodes()
    * return listFromChunkNodes(node.clice(index))
    * ```
-   * 
+   *
    * Note: If index === n, returns an empty list of length 0
    *
    */
@@ -84,7 +84,7 @@ export class ListCompositeTreeViewDU<
     // If slicing from 0, no slicing is necesary
     if (index <= 0) {
       return this;
-    } 
+    }
 
     let newChunksNode;
     let newLength;
--- a/packages/ssz/src/viewDU/listComposite.ts
+++ b/packages/ssz/src/viewDU/listComposite.ts
@@ -1,4 +1,4 @@
-import {Node, treeZeroAfterIndex} from "@chainsafe/persistent-merkle-tree";
+import {Node, subtreeFillToContents, treeZeroAfterIndex, zeroNode} from "@chainsafe/persistent-merkle-tree";
 import {ByteViews, ValueOf} from "../type/abstract";
 import {CompositeType, CompositeView, CompositeViewDU} from "../type/composite";
 import {ListCompositeType} from "../view/listComposite";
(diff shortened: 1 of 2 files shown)


packages/ssz/test/unit/byType/listComposite/tree.test.ts
```diff
@@ -194,3 +194,34 @@ describe("ListCompositeType.sliceTo", () => {
     }
   });
 });
+
+describe("ListCompositeType.sliceFrom", () => {
+  it("Slice List at multiple length", () => {
```
nazarhussain, line 199: Would be better to split the tests into bad and good assertions. e.g. separate test for negative index.

packages/ssz/src/viewDU/listComposite.ts
```diff
@@ -66,6 +66,43 @@ export class ListCompositeTreeViewDU<
...
+
+    // If slicing from 0, no slicing is necesary
+    if (index <= 0) {
+      return this;
+    }
+
+    let newChunksNode;
+    let newLength;
+
+    if (index >= this.nodes.length) {
```
nazarhussain, line 92: I think we should raise error incase user want to slice from out of bound index.

packages/ssz/src/viewDU/listComposite.ts
```diff
@@ -66,6 +66,43 @@ export class ListCompositeTreeViewDU<
...
+   *
+   * ```ts
+   * const nodes = getChunkNodes()
+   * return listFromChunkNodes(node.clice(index))
+   * ```
+   *
+   * Note: If index === n, returns an empty list of length 0
+   *
+   */
+  sliceFrom(index: number): this {
```
nazarhussain, line 80: I feel less confident on the name of the function. `slice` is very common known term and always returns a shallow copy without changing the object. But this implementation is return the current object with mutation. 

May be we call it `sliceAndMerge` or something similar.

packages/ssz/test/unit/byType/listComposite/tree.test.ts
nazarhussain, line 199: I prefer to have detail and atomic test description. e.g. 

> should have correct root hash after slicing

> should have correct sub tree after slicing

nazarhussain (changes requested):
Do we have some specs for this particular behavior?

packages/ssz/src/viewDU/listComposite.ts
tuyennhv: this function returns new ViewDU object without mutating the current ViewDU object
I think the name is good and it conforms to the `sliceTo` method above

ensi321:
> Do we have some specs for this particular behavior? If yes please mention in the PR description.

Updated. Thanks

packages/ssz/src/viewDU/listComposite.ts
ensi321: Yea it is essentially the opposite of `sliceTo`, so naming and behaviour are similar to it.

packages/ssz/src/viewDU/listComposite.ts
ensi321: > I think we should raise error incase user want to slice from out of bound index.

There is a use case for `index === this.nodes.length` to return an empty list. But I agree `index > this.nodes.length` should throw an error 👍

packages/ssz/src/viewDU/listComposite.ts
```diff
@@ -66,6 +66,47 @@ export class ListCompositeTreeViewDU<
     return this.type.getViewDU(newRootNode) as this;
   }
 
+  /**
+   * Returns a new ListCompositeTreeViewDU instance with the values from `index` to the end of list
+   *
+   * ```ts
+   * const nodes = getChunkNodes()
+   * return listFromChunkNodes(node.clice(index))
```
wemeetagain, line 74:
```suggestion
   * return listFromChunkNodes(node.slice(index))
```

packages/ssz/src/viewDU/listComposite.ts
```diff
@@ -66,6 +66,47 @@ export class ListCompositeTreeViewDU<
...
+   *
+   * Note: If index === n, returns an empty list of length 0
+   *
+   */
+  sliceFrom(index: number): this {
+    // Commit before getting rootNode to ensure all pending data is in the rootNode
+    this.commit();
+
+    // If slicing from 0, no slicing is necesary
+    if (index <= 0) {
```
wemeetagain, line 85: from a library perspective, it might be nice to handle negative indices as starting from the end, similarly to how `Array#slice` now works.

```js
[1, 2, 3, 4, 5].slice(-1) // -> [5]
```

should be as simple as adding this above L85:
```ts
if (index < 0) {
  index = length + index
}
```

packages/ssz/src/viewDU/listComposite.ts
```diff
@@ -66,6 +66,47 @@ export class ListCompositeTreeViewDU<
...
+      return this;
+    }
+
+    let newChunksNode;
+    let newLength;
+
+    if (index > this.nodes.length) {
```
wemeetagain, line 92: if we want to mirror `Array#slice` behavior, we don't need this but can just change L96 to `if (index >= this.nodes.length) {`

packages/ssz/test/unit/byType/listComposite/tree.test.ts
```diff
@@ -194,3 +194,65 @@ describe("ListCompositeType.sliceTo", () => {
     }
   });
 });
+
+describe("ListCompositeType.sliceFrom", () => {
+  it("Slice List from -1", () => {
```
wemeetagain, line 199: would like to see tests something like:
```ts
const listType = new ListCompositeType(ssz.Root, 1024);
const listLength = 16;
const list = Array.from({length: listLength}, (_, i) => Buffer.alloc(32, i));
const listView = listType.toViewDU(list);

for (let i = -(listLength + 1); i < listLength + 1; i++) {
  // compare list.slice(i) to listView.sliceFrom(i), they should be equivalent  
  const slicedList = list.slice(i);
  const slicedListView = listView.sliceFrom(i);

  expect(slicedListView.length).to.equal(slicedList.length);
  expect(toHexString(slicedListView.serialize()))
    .to.equal(toHexString(listType.serialize(slicedList)));
  expect(toHexString(slicedListView.hashTreeRoot()))
    .to.equal(toHexString(listType.hashTreeRoot(slicedList)));
}
```
more succinct and covering all cases

wemeetagain (changes requested):
For library usage, it would be nice to also include `sliceFrom` for `ListBasicTreeViewDU`

nazarhussain (dismissed):
Other than @wemeetagain  feedback LGTM.

packages/ssz/src/viewDU/listComposite.ts
```diff
@@ -66,6 +66,47 @@ export class ListCompositeTreeViewDU<
...
+   *
+   * ```ts
+   * const nodes = getChunkNodes()
+   * return listFromChunkNodes(node.slice(index))
+   * ```
+   *
+   * Note: If index === n, returns an empty list of length 0
+   *
+   */
+  sliceFrom(index: number): this {
```
wemeetagain, line 80: can you add a sliceFrom method for ListBasicTreeViewDU ? Or just make an issue for it?

ensi321: > can you add a sliceFrom method for ListBasicTreeViewDU ? Or just make an issue for it?

Is it for the sake of completion? There is no use of `sliceFrom` for `ListBasicTreeViewDU` afaik.

wemeetagain (approved):
Looks good

PR #367, opened 2024-04-15
feat: SIMD implementation for as-sha256

### LINKED ISSUE #356: SIMD implementation for as-sha256
**Is your feature request related to a problem? Please describe.**

With simd supported in assembly script, we can use it to hash multiple inputs in parallel, especially the `digest64()` method

**Describe the solution you'd like**

Use simd instruction in assembly script, we can hash 4 different inputs at the same time which is respective to v128 data structure

**Additional context**

A POC is available in `tuyen/digest_64_simd` branch. Benchmark result:
- on Mac M1 it's 10% - 20% faster
- on a Hetzner node, it's 70% - 80% faster

twoeths:
**Motivation**

SIMD is available in assemblyscript, it supports v128 data structure which mean we can hash 4 inputs in parallel

**Description**

- New assemblyscript simd implementation in `assembly/simd.ts`
- New methods to support hashing 4 inputs (each 64 bytes) in parallel:
  - `hash4Input64s(inputs: Uint8Array[]): Uint8Array[]`
  - `hash8HashObjects(inputs: HashObject[])`
- Add unit tests and benchmarks

Closes #356


tuyennhv:
the performance of simd implementation really depends on the cpu, below is simd vs digest64
- in CI (ubuntu), simd is just a little bit faster
Screenshot 2024-04-19 at 10 17 20

- in my environment (Mac M1) simd is ~20% faster

```
  digest64 vs hash4Input64s vs hash8HashObjects
    ✓ digest64 200092 times                                               6.206878 ops/s    161.1116 ms/op        -         60 runs   10.3 s
    ✓ hash 200092 times using hash4Input64s                               7.460423 ops/s    134.0406 ms/op        -         72 runs   10.2 s
    ✓ hash 200092 times using hash8HashObjects                            7.834839 ops/s    127.6350 ms/op        -         76 runs   10.2 s
```

- in another ubuntu server (which is used for running a lodestar beacon node), simd is almost 2x faster
```
digest64 vs hash4Input64s vs hash8HashObjects
    ✓ digest64 200092 times                                               4.908615 ops/s    203.7235 ms/op        -         47 runs   10.2 s
    ✓ hash 200092 times using hash4Input64s                               9.644699 ops/s    103.6839 ms/op        -         94 runs   10.3 s
    ✓ hash 200092 times using hash8HashObjects                            9.390349 ops/s    106.4923 ms/op        -         90 runs   10.1 s
```

packages/as-sha256/src/index.ts
```diff
@@ -82,6 +76,163 @@ export function digest64HashObjects(obj1: HashObject, obj2: HashObject): HashObj
...
+
+/**
+ * Hash 8 HashObjects in parallel:
+ *   - input${i} has h0 to h7, each 4 bytes which make it 32 bytes
+ *
+ * Inputs      h0    h1    h2    h3    h4    h5    h6   h7
+ *               \   /      \    /       \   /      \   /
+ * Outputs         o0          o1          o2         o3
+ */
+export function hash8HashObjects(inputs: HashObject[]): HashObject[] {
```
g11tech, line 123: i think `8` in name is confusing here because hash objects are 4 only each hash object is h0,...h7

tuyennhv: agree we need to name the 2 new apis better here
- I feel like `digest*()` would be for 1 operation, I propose to make simd apis as `hash*()`
- we have `digest64(data: Uint8Array): Uint8Array`, simd equivalent would be `hash4UintArray64s(inputs: Uint8Array[]): Uint8Array[]` ==> here each input is exactly 64 bytes
- we have `digest64HashObjects(obj1: HashObject, obj2: HashObject): HashObject`, simd equivalent would be `hash4HashObjectInputs(inputs: HashObject[]): HashObject[]` 
  - here `inputs` would be an array of 8 HashObjects (each 2 `HashObjects` is an HashObjectInput). This is easier for the consumer `persistent-merkle-tree` but could be a bit confusing because it's called `hash4*` but input is an array of 8, can add more comments for this
  - other option would be introducing data structure like `{obj1: HashObject; obj2: HashObject}[4]` or `HashObject[2][4]`. This is better in term of clarity but I don't feel like making object creations living really short-tem just for the clarity while the implementation does not need it

let me know what you think @g11tech @wemeetagain

g11tech: i am good with the above suggestion :+1:

g11tech (dismissed):
lgtm

packages/as-sha256/src/index.ts
wemeetagain: Maybe call simd functions with "batch" or "simd" in the name. "Digest" vs "hash" is not great because they are mostly used as synonyms.

tuyennhv: I changed the name to `batchHash*` 👍

packages/as-sha256/assembly/utils/v128.ts
```diff
@@ -0,0 +1,341 @@
...
+    case 60:
+      return v128.load(ptr, 960);
+    case 61:
+      return v128.load(ptr, 976);
+    case 62:
+      return v128.load(ptr, 992);
+    case 63:
+      return v128.load(ptr, 1008);
+    default:
+      return i32x4.splat(0);
```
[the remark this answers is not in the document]
tuyennhv: it's used to create a v128 with 4 32-bit integer lanes are set to 0. I changed the code to throw error instead as this line should not be reached

wemeetagain (approved):

g11tech (approved):

wemeetagain (approved):

PR #372, opened 2024-04-30
chore: bump node version in GitHub workflows

ensi321:
Use node20 and setup-node v4 in GitHub workflows as recommended [here](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/)

This should fix CI errors when attempting to run yarn using node16. See [error log](https://github.com/ChainSafe/ssz/actions/runs/8751637236/job/24030166543)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [20]
+        node: [18, 20]
     steps:
       - uses: actions/checkout@v3
       - run: corepack enable
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,11 +9,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [16, 18]
+        node: [20]
     steps:
       - uses: actions/checkout@v3
       - run: corepack enable
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         with:
           node-version: ${{matrix.node}}
           cache: yarn
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -22,10 +22,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: corepack enable
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         with:
           cache: yarn
-          node-version: 18
+          node-version: 20
       - run: yarn --immutable && yarn build && yarn generate
 
       - name: Run benchmarks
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -9,11 +9,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [16, 18]
+        node: [20]
```
tuyennhv, line 12:
```suggestion
        node: [18, 20]
```

we may still want to support Node 18?

ensi321: > we may still want to support Node 18?

@tuyennhv It looks like in Lodestar we drop all support for Node 18. In particular, Lodestar workflows are all using node20 according to chainsafe/Lodestar#5730. Do we want workflows in ssz to be consistent with Lodestar's?

tuyennhv: as a library I think we should support NodeJS 18 until it's end of life (which is Apr 30 2025)
it could be consumed by different applications, not just lodestar
cc @matthewkeil @nflaig

nflaig: For Lodestar, as in the client, we wanna support the node versions we actually test against, since there is no fleet running v18, it's safer to drop support than have users run into unexpected issues that are hard to debug.

For libraries, we should try to support node versions that are still in active maintenance, including 18 right now.

My idea was that once we start using node 22 that even in the monorepo we still run unit tests against both 20/22. The only tradeoff is that it adds additional load to CI but should be fine for unit.

nflaig: > It looks like in Lodestar we drop all support for Node 18. 

This also doesn't drop support for node 16, unless we added code changes that rely on another minimum node version (which should be communicated in the release notes)

matthewkeil: > as a library I think we should support NodeJS 18 until it's end of life (which is Apr 30 2025) it could be consumed by different applications, not just lodestar cc @matthewkeil @nflaig

Agreed.  We should definitely support 18 for this lib, and when 22 is ready we should add but not drop 18 until 18 has reached end of life

.github/workflows/test.yml
```diff
@@ -9,11 +9,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [16, 18]
+        node: [18, 20]
```
matthewkeil, line 12: Looks good now!  We can add 22 in a couple months once its part of the official releases.  Lets also maintain 18 until end of life.

matthewkeil (approved):
LGTM! 🎸

tuyennhv (approved):

nflaig (approved):

PR #373, opened 2024-05-14
feat: add StableContainer

wemeetagain:
**Motivation**

- https://eips.ethereum.org/EIPS/eip-7495 - SSZ StableContainer
- https://eips.ethereum.org/EIPS/eip-7688 - Forward compatible consensus data structures

**Description**

- Add a `StableContainerType` ssz type that follows the EIP-7495 spec
  - started from a copy/paste of `ContainerType` with several notable diffs
    - optional / inactive fields prevent precomputing offsets, need to be computed on each operation
    - optional field behavior added to every iteration thru `fieldsEntries`
  - sanity unit tests
- Add a `ProfileType` ssz type that follows 'Profile' functionality from the EIP-7495
  - started from a copy/paste of `ContainerType` with several notable diffs
  - there is no linkage between our Profile and its 'upstream' StableContainer. Rather we essentially annotate each field with a chunk index. Currently configured by passing in the active fields bitvector into the type constructor.
  - add a chunkIndex to each field, which lets the implementation merkleize fields in the right place
  - sanity unit tests


packages/ssz/package.tgz
twoeths, line 1: what is this file?

packages/ssz/src/type/stableContainer.ts
```diff
@@ -0,0 +1,794 @@
...
+function isOptionalType(type: Type<unknown>): type is OptionalType<Type<unknown>> {
+  return type instanceof OptionalType;
+}
+
+function toNonOptionalType<T extends Type<unknown>>(type: T): NonOptionalType<T> {
+  return (isOptionalType(type) ? type.elementType : type) as NonOptionalType<T>;
+}
+
+/**
+ * StableContainer: ordered heterogeneous collection of values
```
twoeths, line 75: would be nice to link to the eip

packages/ssz/src/type/stableContainer.ts
```diff
@@ -0,0 +1,794 @@
...
+    }
+
+    return totalSize;
+  }
+
+  value_serializeToBytes(output: ByteViews, offset: number, value: ValueOfFields<Fields>): number {
+    // compute active field bitvector
+    const activeFields = BitArray.fromBoolArray(this.fieldsEntries.map(({fieldName}) => value[fieldName] != null));
+    // write active field bitvector
+    output.uint8Array.set(activeFields.uint8Array, offset);
```
twoeths, line 227: is it safe if `activeFields` is not dividable by 8?

packages/ssz/src/type/stableContainer.ts
```diff
@@ -0,0 +1,794 @@
...
+      totalSize +=
+        fieldType.fixedSize === null ? 4 + fieldType.value_serializedSize(value[fieldName]) : fieldType.fixedSize;
+    }
+
+    return totalSize;
+  }
+
+  value_serializeToBytes(output: ByteViews, offset: number, value: ValueOfFields<Fields>): number {
+    // compute active field bitvector
+    const activeFields = BitArray.fromBoolArray(this.fieldsEntries.map(({fieldName}) => value[fieldName] != null));
```
twoeths, line 225: need to pad to make it a `BitVector`? from the spec: "The Bitvector[N] is padded with False bits up through length N"

packages/ssz/src/view/stableContainer.ts
```diff
@@ -0,0 +1,250 @@
...
+  const isFixedLen: boolean[] = [];
+  const fieldRangesFixedLen: BytesRange[] = [];
+  const variableOffsetsPosition: number[] = [];
+  let pointerFixed = Math.ceil(fields.length / 8);
+
+  for (const [i, {fieldName, fieldType, optional}] of fields.entries()) {
+    // if the field is inactive
+    if (!activeFields.get(i)) {
+      if (!optional) {
+        throw new Error(`Field "${String(fieldName)}" must be active since it is not optional`);
```
twoeths, line 228: looks like latest spec always go with "Optional" fields

packages/ssz/src/view/stableContainer.ts
```diff
@@ -0,0 +1,250 @@
...
+  // Change class name
+  Object.defineProperty(CustomContainerTreeView, "name", {value: type.typeName, writable: false});
+
+  return CustomContainerTreeView as unknown as ContainerTreeViewTypeConstructor<Fields>;
+}
+
+type BytesRange = {start: number; end: number};
+
+/**
+ * Precompute fixed and variable offsets position for faster deserialization.
```
twoeths, line 201: tweak comment for this type nothing that this should be computed once per value instead of once per type like in Container

packages/ssz/src/type/stableContainer.ts
```diff
@@ -0,0 +1,794 @@
...
+
+    return true;
+  }
+
+  /**
+   * Deserializer helper: Returns the bytes ranges of all active fields, both variable and fixed size.
+   * Also returns the active field bitvector.
+   *
+   * Fields may not be contiguous in the serialized bytes, so the returned ranges are [start, end].
+   * - For fixed size fields re-uses the pre-computed values this.fieldRangesFixedLen
```
twoeths, line 524: `fieldRangesFixedLen` is not pre-computed for StableContainer type

packages/ssz/src/type/stableContainer.ts
```diff
@@ -0,0 +1,794 @@
...
+  readonly isViewMutable = true;
+
+  readonly fields: Fields;
+  // Precomputed data for faster serdes
+  readonly fieldsEntries: FieldEntry<NonOptionalFields<Fields>>[];
+  /** End of fixed section of serialized Container */
+  // readonly fixedEnd: number;
+  protected readonly fieldsGindex: Record<keyof Fields, Gindex>;
+  protected readonly jsonKeyToFieldName: Record<string, keyof Fields>;
+  // protected readonly isFixedLen: boolean[];
```
twoeths, line 99: remove these

packages/ssz/src/type/stableContainer.ts
```diff
@@ -0,0 +1,794 @@
...
+  // fast path for depth 1, the bitvector fits in one chunk
+  if (bitLen <= 256) {
+    return new BitArray(rootNode.right.root.subarray(0, Math.ceil(bitLen / 8)), bitLen);
+  }
+
+  const activeFieldsBuf = new Uint8Array(Math.ceil(bitLen / 8));
+  const depth = countToDepth(BigInt(activeFieldsBuf.length));
+  const nodes = getNodesAtDepth(rootNode.right, depth, 0, Math.ceil(bitLen / 256));
+  for (let i = 0; i < nodes.length; i++) {
+    activeFieldsBuf.set(nodes[i].root, i * 32);
```
twoeths, line 717: I don't understand this, `activeFieldsBuf` is allocated as `bitLen / 8` so it's not supposed to store root of a node. Given there are 8 fields, bitLen = 8 so the `activeFieldsBuf` is exactly 1 byte

I suppose each bit is respective to 1 field, so if node is `Bytes32()` it's a `true`, otherwise it's a `false`?

packages/ssz/test/unit/byType/stableContainer/valid.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      root: "0xd585dd0561c718bf4c29e4c1bd7d4efd4a5fe3c45942a7f778acb78fd0b2a4d2",
+    },
+    {
+      id: "shape2-4",
+      serialized: "0x060600000042000102",
+      json: {side: null, colors: [1, 2], radius: 0x42},
+      root: "0x00fc0cecc200a415a07372d5d5b8bc7ce49f52504ed3da0336f80a26d811c7bf",
+    },
+  ],
+});
```
twoeths, line 132: would be great to add a test case simulating adding new fields after a hard fork in ethereum:

```typescript
const Type1 = new StableContainerType(
  {
    a: optionalUint16,
    b: optionalUint16,
  },
  4
);

const Type2 = new StableContainerType(
  {
    a: optionalUint16,
    b: optionalUint16,
    c: optionalUint16
  },
  4
);
```

then given a value `x` of `Type1`, we serialize it to `data`, then we can use `Type2` to deserialize that `data` to value `x` again. Also the `hashTreeRoot()` of x should be the same using both `Type1` and `Type2`

packages/ssz/src/type/stableContainer.ts
twoeths: should be safe

packages/ssz/src/type/stableContainer.ts
twoeths: this is supposed to not to get to the data nodes, but the activeFileds nodes

```
const depth = countToDepth(BigInt(activeFieldsBuf.length));const depth = countToDepth(BigInt(activeFieldsBuf.length));
```

packages/ssz/src/view/stableContainer.ts
twoeths: this implementation supports both Optional/NonOptional which is great

packages/ssz/src/type/stableContainer.ts
twoeths: `BitArray` is only the value not type. Still follow BitVector spec there

twoeths:
@wemeetagain a summary of the changes I made:
- refactor to have `BasicContainerTreeViewDU` on our current container viewdu
- `StableContainerTreeViewDU` and `ProfileTreeViewDU` to inherit from `BasicContainerTreeViewDU` so that they support `batchHashTreeRoot`
- go with latest spec: support OptionalType for Profile, prepend BitVector[N] to StableContainer (a bug fix)
- add a lot of tests for both types
  - backward compatible test for StableContainer and cross test merkleization Profile vs StableContainer

twoeths (approved):

PR #380, opened 2024-07-11
fix: do not rebind nodes if child ViewDU is not changed

### LINKED ISSUE #379: ViewDU container rebinds node with read-only operations
**Describe the bug**

Consider this unit test here in #378 

```typescript
  it("should not recompute hashTreeRoot() when no fields is changed", () => {
    const root = state.hashTreeRoot();
    // this causes viewsChanged inside BeaconState container
    state.validators.length;
    state.balances.length;
    // but we should not recompute root, should get from cache instead
    const root2 = state.hashTreeRoot();
    expect(root2).to.equal(root, "should not recompute hashTreeRoot() when no fields are changed");
  });
```
the BeaconState container tracked `validators` and `length` as `"viewsChanged"` and it rebinds the underlying tree to compute new root, which results in the same value but different tree/nodes

note that #378 implemented a cache that returns the same root for multiple `ViewDU.hashTreeRoot()` calls

**Expected behavior**

- according to the design of ssz v2, BeaconState container should still track `validators` and `balances` as "viewsChanged" so that when we modify them they will escalate the change to BeaconState. However when we commit those viewDUs, if we see no change we should not rebind the nodes, and it should return the same cached root

**Steps to Reproduce**

- Run above unit test on #378

twoeths:
**Motivation**

There are some read-only operations during state transition, for example: state.balances.length, state.validators.length, state.latestBlockHeaders

although the data is not changed, ViewDU still track it as changed node and blindly rebind nodes, which takes time to rebind and compute roots

**Description**

- compare to the current tracked nodes before doing the update, this is for composite-type related super type
Closes #379

### DIFF
--- a/packages/ssz/src/viewDU/abstract.ts
+++ b/packages/ssz/src/viewDU/abstract.ts
@@ -67,7 +67,7 @@ export abstract class TreeViewDU<T extends CompositeType<unknown, unknown, unkno
     // remember not to do a commit() before calling this function
     // in ethereum consensus, the only type goes with TVDU is BeaconState and it's really more efficient to hash the tree in batch
     // if consumers don't want to batch hash, just go with `this.node.root` similar to what View.hashTreeRoot() does
-
+    // there should not be another ViewDU.hashTreeRoot() during this flow so it's safe to reuse nextHashComps
     const hashComps = nextHashComps;
     this.commit(hashComps);
     if (nextHashComps.byLevel.length > 0 || nextHashComps.offset !== 0) {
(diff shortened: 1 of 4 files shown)


twoeths:
commit time is not worse than `te/batch_hash_tree_root` (actually it's ~ 2ms faster)

- this branch
Screenshot 2024-07-11 at 10 04 16

-  `te/batch_hash_tree_root`
Screenshot 2024-07-11 at 10 04 41

`hashTreeRoot()` is ~10ms faster due to less hash computations, but this `nogroup` node has no validator vs `feat1-1k`

PR #378, opened 2024-06-16
Te/batch hash tree root

matthewkeil:
**Motivation**

__Created on behalf of @twoeths.  Keeping in draft for now but using to store some metrics from different iterations of the changes.__

Update system to allow for batch hashing of trees.  SIMD instruction set allows for parallel processing of hashes when running on a single core. Add the idea of levels to the hashing process to allow deeper sections of a tree to be processed before higher levels and then computes the hashes according to depth of the tree.

PR #382, opened 2024-07-14
feat!: implement hashInto() api for as-sha256

twoeths:
**Motivation**

The hashInto() api of hashtree is very efficient in terms of memory allocation for the consumer because they can preallocate output and reuse over multiple hashes, they can also hash into the same input Uint8Array

**Description**

- implement hashInto() api with the same interface to hashtree
- implement `digest64HashObjectsInto()` which is similar to `digest64HashObjects()` but this does not allocate new HashObject
- implement `byteArrayIntoHashObject()` which is similar to `byteArrayToHashObject()`, but this does not allocate new HashObject

all of these functions are used by `persistent-merkle-tree` and `ssz` in upcoming PRs

**BREAKING CHANGE**
- add `offset` param to `byteArrayToHashObject()` so that consumer can use the same big Uint8Array to extract HashObject from a subsection of it without using `subarray()`

part of #378

### DIFF
--- a/packages/as-sha256/src/hashObject.ts
+++ b/packages/as-sha256/src/hashObject.ts
@@ -98,95 +98,103 @@ export function hashObjectToByteArray(obj: HashObject, byteArr: Uint8Array, offs
  * This function contains multiple same procedures but we intentionally
  * do it step by step to improve performance a bit.
  **/
-export function byteArrayToHashObject(byteArr: Uint8Array): HashObject {
-  let tmp = 0;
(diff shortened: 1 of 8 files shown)


wemeetagain (approved):
Can you add a section to the PR description labeled BREAKING CHANGE that describes what has broken?

twoeths:
> Can you add a section to the PR description labeled BREAKING CHANGE that describes what has broken?

I just added it 👍

g11tech (approved):
lgtm post merge review

Issue #355, opened 2024-03-20
Investigating Tree Hashing in Lodestar


wemeetagain:
## Current Hashing Approach

Lodestar's architecture relies heavily on maintaining a full merkle tree of 
the beacon state. We represent the tree as a linked data structure, where each node is 1. immutable, 2. lazily computing but caching the hash of its children.

This allows us to minimize the number of hashes we need to perform during state transition. Since all hashes of a prestate are maintained, only the paths thru the tree to the "diff" need to be re-hashed. This reduces the computational cost of hashing.

Also, it allows us perform _structural sharing_, sharing the memory for beacon states with shared subtrees. For example, between epochs in a sync period, where sync committees remain constant, if we maintain a reference to two beacon states, both states will share the same underlying subtrees for the sync committees. This reduces the memory cost of maintaining several related states.

### Hashing Function

We use as-sha256 with several optimizations
- we rely on hash inputs always being 64 bytes - this allows us to precompute part of the sha256 internals for a decent gain
    - https://github.com/ChainSafe/as-sha256/pull/51
    - https://github.com/ChainSafe/as-sha256/pull/59
- we avoid allocations inside library, only using fixed input/output buffers

- Related hashing perf analysis https://github.com/ChainSafe/lodestar/issues/2206

### Hash Cache Representation

The memory usage for each type of object in Javascript is not very efficient coming from a systems language intuition. We store hash objects NOT as 32-byte Uint8Array. A 32-byte Uint8Array takes 223 total bytes! There's a bunch of pointers and additional bookkeeping that's being stored behind the scenes.
We store hash objects as objects with 8 uint32 numbers. eg: `{h0: 0, h1: 0, ..., h7: 0}`. This takes somewhere between 88 bytes and 216 bytes, depending on the sizes of the indiviudal component numbers. Smaller numbers are represented as Smi (small integer) as an immediate value, while larger numbers are stored on the heap. In practice, this happens TODO.

- https://github.com/ChainSafe/as-sha256/pull/58
- https://github.com/ChainSafe/lodestar/issues/2885
- https://github.com/ChainSafe/ssz/pull/219

## How to improve?

The hashing speed in lodestar is quite low compared to other implementations.

- Investigate hashtree implementation, which hashes multiple hashes at once. 
- Investigate using a systems language implementation, eg using napi-rs. This could be tried using popular libraries, which likely attempt some hardware acceleration. Also can be tried using a transliteration of as-sha256.

[...]

PR #384, opened 2024-07-16
feat: implement batch hash utils

twoeths:
**Motivation**

- Implement batch hash utils to be consumed by ssz in the upcoming PRs
- cherry-picked from #378

**Description**

- consume `@chainsafe/hashtree`, see `hashtree.ts`
- Model `HashComputations` to compute a parent hash from 2 child nodes
  - getHashComputations() to populate HashComputations[][]
  - executeHashComputations() execute HashComputations from bottom nodes up until the root node, this is part of hasher api
- other new methods for hashers:
  - merkleizeInto: merkleize() without memory allocation, pad `zeroHash[i]` in the middle if needed
  - digestNLevel: hash every 64 bytes up until some level, used when we want to hash multiple validators at the same time (later)
- modify `setNodesAtDepth()` function by adding an output param to compute HashComputations, this saves one traversal from the consumer

part of #355


packages/persistent-merkle-tree/src/node.ts
```diff
@@ -1,8 +1,21 @@
...
-import {hashObjectToUint8Array, hasher, uint8ArrayToHashObject} from "./hasher";
+import {executeHashComputations, hashObjectToUint8Array, hasher, uint8ArrayToHashObject} from "./hasher";
 
 const TWO_POWER_32 = 2 ** 32;
 
+export type HashComputation = {
+  src0: Node;
+  src1: Node;
+  dest: Node;
+};
```
twoeths, line 10: not sure this is the best name, other name could be `{left: Node; right: Node; parent: Node}`

g11tech (approved):
looks amazing!! great work Tuyen! :rocket:

wemeetagain (approved):

PR #385, opened 2024-07-18
feat: batch hash ViewDU

twoeths:
**Motivation**

- BeaconState in lodestar is modeled as a ViewDU, implement batch hash to speed up `hashTreeRoot()` time there
- part of #378 

**Description**
- In `TreeViewDU.hashTreeRoot()`, during `commit()` phase also compute HashComputations using `setNodesAtDepth()` in the previous PR #384, then finally call `executeHashComputations()` in the previous PR #384
  - add optional parameter `hashComps?: HashComputationGroup | null` to `ViewDU.commit()` method. If calling from `hashTreeRoot()`, pass a `HashComputationGroup` there, else null
- Right now lodestar validator is modeled as a ContainerNodeStructViewDU that make it impossible to compute HashComputations there because it's not tree backed. So I came up a solution to also compute validator roots in batch in `ListValidatorTreeViewDU.commit()`
  - one key improvement is not to create validator tree (which takes a lot of memory) when computing its root
  - computing (ContainerNodeStruct) validators in `commit()` seems a bit abnormal but:
    - it does not break the current design: we still can change data multiple times before the commit(), we can also do multiple commit() calls
    - we always have to do `hashTreeRoot()` per state transition
- benchmark in `packages/ssz/test/perf/eth2/beaconState.test.ts` showed ~7x improvement in my Mac M1 compared to the previous release

**Implementation Notes for lodestar**
- need to sync `ListValidatorTreeViewDU` and `ListValidatorType` and define validators as this new type there

**Note for reviewers**
- start from `packages/ssz/src/viewDU/abstract.ts`, `hashTreeRoot()` method
- review `packages/ssz/test/lodestarTypes/phase0/viewDU/listValidator.ts` on the way to hash multiple validators in batch using single memory allocation

### DIFF
--- a/packages/ssz/src/type/arrayBasic.ts
+++ b/packages/ssz/src/type/arrayBasic.ts
@@ -1,10 +1,12 @@
 import {
   BranchNode,
+  HashComputationGroup,
   LeafNode,
   Node,
   getNodesAtDepth,
   packedNodeRootsToBytes,
   packedRootsBytesToNode,
+  arrayAtIndex,
 } from "@chainsafe/persistent-merkle-tree";
 import {Type, ValueOf, ByteViews} from "./abstract";
 import {BasicType} from "./basic";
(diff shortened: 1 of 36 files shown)


twoeths:
due to #386 a lot of apis are refactored so closing this PR for now

PR #388, opened 2024-07-23
feat: minimal memory allocation on ViewDU hashTreeRoot

twoeths:
**Motivation**

Right now HashComputationGroup is allocated every time we call `ViewDU.hashTreeRoot()`, we need design the api to allow consumer to preallocate it to save more memory allocation

see #386

**Description**

- refactor `HashComputationLevel` to use LinkedList, this will allocate minimal memory when it wants to grow
- do not allocate temporary array in `hashtree.ts`
- do not allocate temporary objects in `ViewDU.commit()`

Closes #386

Issue #386, opened 2024-07-19
Minimal memory allocation for ViewDU batch hash


twoeths:
**Is your feature request related to a problem? Please describe.**

batch hash work is in progress, this is how it looks in a lodestar node every time we get through epoch transition

- 6h chart

<img width="1326" alt="Screenshot 2024-07-19 at 08 18 18" src="https://github.com/user-attachments/assets/902676e3-7667-431e-bcf1-2806da30bafe">

- 10m chart

<img width="1324" alt="Screenshot 2024-07-19 at 08 23 36" src="https://github.com/user-attachments/assets/8c6d0ff5-d29e-43d6-b43f-fb8ef0978a85">

looks like memory allocation is an issue there and it caused some spiked `hash_tree_root` times

**Describe the solution you'd like**

right now we allocate memory every time we compute `ViewDU.hashTreeRoot()`, find a way to do minimal memory allocation and reuse it over time

**Additional context**

- tested branch: `te/batch_hash_tree_root`

PR #389, opened 2024-07-23
feat: implement HashComputationLevel using LinkedList

twoeths:
**Motivation**

- memory takes time when testing batch hash, see #386 
- need to design `HashComputationGroup` and `HashComputationLevel` to be able to reuse at the consumer side where it makes sense (for example: processBlocks, compute root for BeaconState.balances)
  - modify the previous implementation: #384

**Description**
- Move `HashComputation` type and its utils to a separate class
- Implement `HashComputationLevel` to be a linked-list like to save memory allocation:
  - when push() may reuse the existing item of LinkedList instead of allocate one
  - when it grows it only need to add one more list items, instead of allocating a whole new array to reuse and save memory
  - loop speed is same to a regular array
  - we don't need random access
- others:
  - `hashtree`: allocate `destNodes` variable once, set it as default
  - `subtreeFillToContents` + `setNodesAtDepth`: destructure params in order not to allocate temporary objects at the consumer side

test result was persisted [here](https://github.com/ChainSafe/ssz/pull/388#issuecomment-2244446980)

part of #378 

cc @wemeetagain @g11tech


g11tech:
> when it grows it only need to add one more list items, instead of allocating a whole new array to reuse and save memory

this is where it optimizes memory?

packages/persistent-merkle-tree/src/hashComputation.ts
```diff
@@ -0,0 +1,216 @@
...
+  push(src0: Node, src1: Node, dest: Node): void {
+    if (this.tail !== null) {
+      let newTail = this.tail.next;
+      if (newTail !== null) {
+        newTail.src0 = src0;
+        newTail.src1 = src1;
+        newTail.dest = dest;
+      } else {
+        // grow the list
+        newTail = {src0, src1, dest, next: null};
```
g11tech, line 81: so this part optimizes memory over the array implementation?

twoeths: yes, when we are at length 1M, for example, pushing a new item cause very small allocation for LinkedList while could be a big deal for an array, especially array of objects

that's why we see a lot of 1s spikes using Array

wemeetagain: As far as I know, v8 `Array` is like an object, which shouldn't incur a large allocation when new properties are added.

So I'm not really sure why this works from a theory perspective, but the data looks promising, and can't argue with that :shrug:

packages/persistent-merkle-tree/src/hashComputation.ts
```diff
@@ -0,0 +1,216 @@
...
+  dest: Node;
+  next: HashComputation | null;
+};
+
+/**
+ * Model HashComputation[] at the same level that support reusing the same memory.
+ * Before every run, reset() should be called.
+ * After every run, clean() should be called.
+ */
+export class HashComputationLevel {
```
wemeetagain, line 24: can you add `implements IterableIterator<HashComputation>` here?

packages/persistent-merkle-tree/src/hashComputation.ts
```diff
@@ -0,0 +1,216 @@
...
+    for (const hc of this) {
+      hashComps.push(hc);
+    }
+    return hashComps;
+  }
+
+  /**
+   * For testing.
+   */
+  dump(): HashComputation[] {
```
wemeetagain, line 160: Is this also needed? seems that either `toArray` or `dump` should exist

packages/persistent-merkle-tree/src/hashComputation.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  clean(): void {
+    for (const level of this.byLevel) {
+      level.clean();
+    }
+  }
+}
+
+/**
+ * Get HashComputations from a root node all the way to the leaf nodes.
```
wemeetagain, line 193: can you mention what is mutated / updated here?

packages/persistent-merkle-tree/src/hashComputation.ts
```diff
@@ -0,0 +1,216 @@
...
+ */
+export function getHashComputations(node: Node, offset: number, hcByLevel: HashComputationLevel[]): void {
```
wemeetagain, line 195: maybe name `offset` as `levelIx` or `index`

packages/persistent-merkle-tree/src/hashComputation.ts
twoeths: yes I tried to answer the same question, looks like when adding a new object to Array, if it reaches its internal memory it needs to allocate and copy all items to a new space. This response from ChatGPT looks reasonable to me:
```
JavaScript arrays are dynamically sized, but internally, they need to manage memory blocks to store the array elements:

Allocation on Growth: If the existing capacity of the array is insufficient to accommodate a new element, the JavaScript engine must allocate a new, larger block of memory to hold the array's contents and copy existing elements to this new space. This operation increases temporary memory usage and may trigger GC if the memory is tight.
Fragmentation: Repeatedly resizing arrays can lead to memory fragmentation, which might prompt the garbage collector to run more frequently to reclaim unused memory segments.
```

packages/persistent-merkle-tree/src/hashComputation.ts
twoeths: `toArray()` will return all HashComputation with data while `dump()` returns that + those with no data (some `HashComputation` actually only a place holder for the next item to be added)

I updated the comment for those methods

wemeetagain (approved):

PR #390, opened 2024-07-31
feat: ReuseListIterator for getAll() api

twoeths:
**Motivation**

- avoid possible array allocation with getAll() api

#383 make it possible to provide an Array for getAll() api, however it could still spike due to array allocation

**Description**
- Implement ReusableListIterator which is a LinkedList, when it grows it only append some more items instead of allocating a big array
- Implement getAll*Iter()  api to View and ViewDU to allow consumer to use `ReusableListIterator`
- On average, ReusableListIterator is only a bit faster than Array however the key thing is to reduce any spikes as much as possible due to gc

**Testing**
- This branch (lg1k) in the last 24h, 10m rate interval, there are ~3 spikes

Screenshot 2024-07-31 at 11 20 18

- `te/batch_hash_tree_root`, md64, 10m rate interval, there are ~18 spikes

Screenshot 2024-07-31 at 11 20 59

- This branch (lg1k) in the last 24h, 12h rate interval. It's only a little bit faster than `te/batch_hash_tree_root`

Screenshot 2024-07-31 at 11 24 35

- `te/batch_hash_tree_root`, md64, 12h rate interval

Screenshot 2024-07-31 at 11 25 16

### DIFF
--- a/packages/persistent-merkle-tree/src/hashComputation.ts
+++ b/packages/persistent-merkle-tree/src/hashComputation.ts
@@ -55,7 +55,10 @@ export class HashComputationLevel {
    * run before every run
    */
   reset(): void {
-    // keep this.head
+    // keep this.head object, only release the data
+    this.head.src0 = null as unknown as Node;
+    this.head.src1 = null as unknown as Node;
+    this.head.dest = null as unknown as Node;
     this.tail = null;
     this._length = 0;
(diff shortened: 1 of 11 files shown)


twoeths:
note that there are still spikes in lodestar, see https://github.com/ChainSafe/lodestar/pull/6979

PR #391, opened 2024-08-05
feat: use allocUnsafe to allocate hash digests

wemeetagain:
**Motivation**

- allocation is typically slow, using allocUnsafe to allocate hash digests can help

**Description**

- use allocUnsafe where possible to allocate hash digests
- use subarray to avoid an additional data copy when using digest with long inputs.


twoeths (approved):
looks good to me, I think we should apply `alloc` utils to `ssz hashTreeRoot()` too

wemeetagain:
> we should apply alloc utils to ssz hashTreeRoot() too

looks like the ssz lib will be using the digest functions that have been updated here

PR #392, opened 2024-08-09
feat: implement ViewDU.batchHashTreeRoot()

twoeths:
**Motivation**

- Implement `batchHashTreeRoot()` for ViewDU type based on `HashComputation`

**Description**
- `hashTreeRoot()` is still great because it does not allocate any memory so we need a new method to compute batch
- in `packages/ssz/src/viewDU/abstract.ts`, add `batchHashTreeRoot()` method
  - param is `HashComputationGroup` in previous PR [389](https://github.com/ChainSafe/ssz/pull/389/files#diff-60bb8e753e65c883c3183feb604cb9e9f9a1ae2c5be4aa0f0add8706ca7c907cR174)
  - param is designed to be reused
- add new params to `commit()` method:
  - hcOffset: offset from the top level
  - hcByLevel: an array of HashComputationLevel[] or null. 
  - when `hcByLevel` is null, it means it was called from `hashTreeRoot()`, `commit()` method should be backward compatible in this case
  - ContainerNodeStructTreeViewDU:
    - when commit(), it's unchanged
    - if there is a not null `hcByLevel`, it should compute root because this ViewDU does not support `HashComputation`
 
cherry picked from #378

**Unit Tests**
- I added a bunch of tests to make sure `batchHashTreeRoot()` returns the same value to `hashTreeRoot`, please suggest if we need more

part of #355

### DIFF
--- a/packages/persistent-merkle-tree/test/perf/hasher.test.ts
+++ b/packages/persistent-merkle-tree/test/perf/hasher.test.ts
@@ -79,7 +79,7 @@ describe("hashtree", function () {
   });
 
   itBench({
-    id: `executeHashComputations - hashtree`,
+    id: `executeHashComputations`,
     beforeEach: () => {
       const [tree] = buildComparisonTrees(16);
       return tree;
@@ -92,7 +92,7 @@ describe("hashtree", function () {
   });
 
   itBench({
-    id: `root - hashtree`,
+    id: `get root`,
     beforeEach: () => {
       const [tree] = buildComparisonTrees(16);
       setHasher(hashtreeHasher);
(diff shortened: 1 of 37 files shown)


twoeths:
I should have mentioned the new `ViewDU.batchHashTreeRoot()` does not affect `hashTreeRoot()`, that's one of the criterias to merge these PRs

wemeetagain (approved):

twoeths:
this is part 1 of `ViewDU.batchHashTreeRoot()`, part 2 is to batch hash at `commit()` for ListValidator which requires the new `type.hashTreeRootInto()` api which comes next

PR #394, opened 2024-08-19
fix: sliceFrom api

twoeths:
**Motivation**

- sliceFrom() api was implemented with the assumption that `this.nodes` are fully populated but it's not unless `populateAllNodes()` is called

**Description**

- reproduce the bug in unit test
- call `populateAllNodes()` api to fix it


nflaig (approved):

wemeetagain (approved):

twoeths:
the issue did not happen in lodestar because we always get all nodes before calling `sliceFrom` https://github.com/ChainSafe/lodestar/blob/d69d809e7ab5d5a2cf5030429e83b6af1d5af31a/packages/state-transition/src/epoch/processPendingBalanceDeposits.ts#L24

PR #405, opened 2024-10-01
feat: new api to upgrade a tree to new type

twoeths:
**Motivation**

Cannot reuse an existing tree when upgrading to new type, see https://github.com/ChainSafe/lodestar/issues/6811

**Description**

- new "upgradeToNewType" api as discussed [Lodestar#6811](https://github.com/ChainSafe/lodestar/issues/6811#issuecomment-2129571320)

- usage:

```typescript
  const newType = new ContainerType(newFields);
  const newView = newType.getView(new Tree(view.upgradeToNewType(newType)));
  const newViewDU = newType.getViewDU(viewDU.upgradeToNewType(newType));
```


wemeetagain (approved):
LGTM, I refactored slightly to _only_ export the utility function `upgradeToNewType`

PR #409, opened 2024-10-15
feat: improve type.hashTreeRoot() using batch

twoeths:
**Motivation**

- improve type.hashTreeRoot() using batch

**Description**
- instead of `getRoots()` and compute root from there, this PR implement `getChunkBytes()`
  - this compute root using `merkleizeInto()` which use batch there
  - reuse `chunkBytesBuffer` memory in type,  almost no `Uint8Array` allocations in the middle
- new `hashTreeRootInto()` api. This is needed in case consumers want to reuse memory allocation there
- use `allocUnsafe()` of as-sha256 where it makes sense

cherry picked from #378

### DIFF
--- a/packages/ssz/src/type/basic.ts
+++ b/packages/ssz/src/type/basic.ts
@@ -30,11 +30,18 @@ export abstract class BasicType<V> extends Type<V> {
   }
 
   hashTreeRoot(value: V): Uint8Array {
-    // TODO: Optimize
-    const uint8Array = new Uint8Array(32);
+    // cannot use allocUnsafe() here because hashTreeRootInto() may not fill the whole 32 bytes
+    const root = new Uint8Array(32);
+    this.hashTreeRootInto(value, root, 0);
+    return root;
+  }
+
+  hashTreeRootInto(value: V, output: Uint8Array, offset: number): void {
+    const uint8Array = output.subarray(offset, offset + 32);
+    // output could have preallocated data, some types may not fill the whole 32 bytes
+    uint8Array.fill(0);
     const dataView = new DataView(uint8Array.buffer, uint8Array.byteOffset, uint8Array.byteLength);
     this.value_serializeToBytes({uint8Array, dataView}, 0, value);
-    return uint8Array;
   }
 
   clone(value: V): V {
--- a/packages/ssz/src/type/byteArray.ts
+++ b/packages/ssz/src/type/byteArray.ts
@@ -8,7 +8,6 @@ import {
   getHashComputations,
 } from "@chainsafe/persistent-merkle-tree";
 import {fromHexString, toHexString, byteArrayEquals} from "../util/byteArray";
-import {splitIntoRootChunks} from "../util/merkleize";
 import {ByteViews} from "./abstract";
 import {CompositeType, LENGTH_GINDEX} from "./composite";
 
@@ -82,10 +81,23 @@ export abstract class ByteArrayType extends CompositeType<ByteArray, ByteArray, 
     return Uint8Array.prototype.slice.call(data.uint8Array, start, end);
   }
 
+  value_toTree(value: ByteArray): Node {
+    // this saves 1 allocation of Uint8Array
+    const dataView = new DataView(value.buffer, value.byteOffset, value.byteLength);
+    return this.tree_deserializeFromBytes({uint8Array: value, dataView}, 0, value.length);
+  }
+
   // Merkleization
 
-  protected getRoots(value: ByteArray): Uint8Array[] {
-    return splitIntoRootChunks(value);
+  protected getChunkBytes(value: ByteArray): Uint8Array {
+    // reallocate this.merkleBytes if needed
+    if (value.length > this.chunkBytesBuffer.length) {
+      const chunkCount = Math.ceil(value.length / 32);
+      const chunkBytes = chunkCount * 32;
+      // pad 1 chunk if maxChunkCount is not even
+      this.chunkBytesBuffer = chunkCount % 2 === 1 ? new Uint8Array(chunkBytes + 32) : new Uint8Array(chunkBytes);
+    }
+    return getChunkBytes(value, this.chunkBytesBuffer);
   }
 
   // Proofs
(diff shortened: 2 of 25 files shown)


twoeths:
tested this on `feat1`, see https://github.com/ChainSafe/lodestar/pull/7171#issuecomment-2425472397
ready to review

twoeths:
sha256 works in blocks, each is 64 bytes so perhaps it's more meaningful to reflect that for `chunkBytesBuffer` variable

also with holesky, there are 1.7M validators. For every 8 deposits we have to reallocate the whole 1.7M * 8 bytes = 13.6MB for `BeaconState.balances` which is not ideal. Need to instead allocate another 64 bytes in this case. This applies for all list types.

packages/ssz/src/type/byteArray.ts
```diff
@@ -149,3 +159,16 @@ export abstract class ByteArrayType extends CompositeType<ByteArray, ByteArray,
...
 }
+
+export function getBlocksBytes(value: Uint8Array, blocksBuffer: Uint8Array): Uint8Array {
+  if (value.length > blocksBuffer.length) {
+    throw new Error(`data length ${value.length} exceeds blocksBuffer length ${blocksBuffer.length}`);
+  }
+
+  blocksBuffer.set(value);
+  const valueLen = value.length;
+  const blockByteLen = Math.ceil(valueLen / 64) * 64;
```
matthewkeil, line 170: It might be helpful to break out a helper function so its clear why this is happening everywhere.

```ts
export function getPaddedByte32Count(buf: ArrayBuffer): number {
    return Math.ceil(buf.length / 32);
}

export function getPaddedByte64Count(buf: ArrayBuffer): number {
    return Math.ceil(buf.length / 64);
}
```

packages/ssz/src/type/byteArray.ts
```diff
@@ -82,10 +81,21 @@ export abstract class ByteArrayType extends CompositeType<ByteArray, ByteArray,
...
+    // this saves 1 allocation of Uint8Array
+    const dataView = new DataView(value.buffer, value.byteOffset, value.byteLength);
+    return this.tree_deserializeFromBytes({uint8Array: value, dataView}, 0, value.length);
+  }
+
   // Merkleization
 
-  protected getRoots(value: ByteArray): Uint8Array[] {
-    return splitIntoRootChunks(value);
+  protected getBlocksBytes(value: ByteArray): Uint8Array {
```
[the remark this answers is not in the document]
matthewkeil: Perhaps calling 32 byte chunks `Bytes32` and 64-byte "blocks" `Bytes64`?  Not sure, what do you think?

packages/ssz/src/type/byteList.ts
```diff
@@ -89,7 +104,49 @@ export class ByteListType extends ByteArrayType {
...
+    // use  merkleizeBlockArray() instead of merkleizeBlocksBytes() to avoid big memory allocation
+    // reallocate this.blockArray if needed
+    if (value.length > this.blockBytesLen) {
+      const newBlockCount = Math.ceil(value.length / 64);
+      // this.blockBytesLen should be a multiple of 64
+      const oldBlockCount = Math.ceil(this.blockBytesLen / 64);
+      const blockDiff = newBlockCount - oldBlockCount;
+      const newBlocksBytes = new Uint8Array(blockDiff * 64);
+      for (let i = 0; i < blockDiff; i++) {
+        this.blockArray.push(newBlocksBytes.subarray(i * 64, (i + 1) * 64));
```
matthewkeil, line 126: Why loop and push in chunks?  Why not just push all at once and only trigger one resize to the `blockArray`?

matthewkeil (commented):
I left a few comments but I think this PR really needs to be reviewed by @wemeetagain

packages/ssz/src/type/byteList.ts
nazarhussain: The `blockArray` is an array of Uint8Array, we need to push 64 byte chunks as elements. a.k.a. Linear bytes to array of bytes.

packages/ssz/src/type/byteArray.ts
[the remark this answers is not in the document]
nazarhussain: @matthewkeil I am in favor of renaming this function to `getPaddedBytes64`.  But I feel it should be done separately across the repo. 

A **Block** refers to 64 byte chunk in SHA256, and we are using this term a lot on other places as well. So if we refactor this term should be done across all usages, which may also result in API interface changes of the package. 

https://github.com/ChainSafe/ssz/blob/62e9a137c7a6a405daca61af98eb08b93f774d95/packages/persistent-merkle-tree/src/hasher/types.ts#L22

https://github.com/ChainSafe/ssz/blob/62e9a137c7a6a405daca61af98eb08b93f774d95/packages/persistent-merkle-tree/src/hasher/types.ts#L29

https://github.com/ChainSafe/ssz/blob/62e9a137c7a6a405daca61af98eb08b93f774d95/packages/ssz/src/type/byteList.ts#L44

Suggest what do you prefer now?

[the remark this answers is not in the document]
matthewkeil: maybe leaving it like it is for now is the right idea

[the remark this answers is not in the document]
nazarhussain: Created an issue to do that change in a separate PR. https://github.com/ChainSafe/ssz/issues/468

packages/ssz/src/type/basic.ts
```diff
@@ -30,11 +30,18 @@ export abstract class BasicType<V> extends Type<V> {
   }
 
   hashTreeRoot(value: V): Uint8Array {
-    // TODO: Optimize
-    const uint8Array = new Uint8Array(32);
+    // cannot use allocUnsafe() here because hashTreeRootInto() may not fill the whole 32 bytes
```
wemeetagain, line 33: is this comment correct? it looks like hashTreeRootInto is filling all 32 bytes

wemeetagain (approved):

packages/ssz/src/type/basic.ts
twoeths: I'd remove that comment @nazarhussain . I remember I tried `allocUnsafe()` at some points and got "Invalid state root" issue. If we want to apply it, need to audit carefully an apply to other places as well

wemeetagain (approved):

nazarhussain (approved):

PR #417, opened 2024-10-21
feat: getAll() apis to support output parameters

### LINKED ISSUE #379: ViewDU container rebinds node with read-only operations

twoeths:
**Motivation**

- right now at lodestar, it has to allocate new array to loop through all elements and it cost memory allocation
- for readonly operation, ssz still have to recompute root, see #379

**Description**

- improve getAll() apis by supporting output parameters
- add `forEach()` apis
- if a ViewDU does not change root node after a commit(), its parent should discard the change and so that it does not have to recompute root

Closes #379

### DIFF
--- a/packages/ssz/src/view/arrayBasic.ts
+++ b/packages/ssz/src/view/arrayBasic.ts
@@ -87,14 +87,18 @@ export class ArrayBasicTreeView<ElementType extends BasicType<unknown>> extends 
 
   /**
    * Get all values of this array as Basic element type values, from index zero to `this.length - 1`
+   * @param values optional output parameter, if is provided it must be an array of the same length as this array
    */
-  getAll(): ValueOf<ElementType>[] {
+  getAll(values?: ValueOf<ElementType>[]): ValueOf<ElementType>[] {
+    if (values && values.length !== this.length) {
+      throw Error(`Expected ${this.length} values, got ${values.length}`);
+    }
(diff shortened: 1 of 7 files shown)


twoeths:
this was tested in lodestar, see https://github.com/ChainSafe/lodestar/pull/7171#issuecomment-2425472397

PR #421, opened 2024-11-06
feat: implement merkleizeBlockArray

twoeths:
**Motivation**

- when we have a list of bytes or list of composite, if new value is over the cached chunk/block bytes, right now we reallocate the whole Uint8Array which is not good for gc

**Description**

- Instead of that, implement `merkleizeBlockArray(blocks: Uint8Array[], ...)`. In the above case we only need to allocate some more Uint8Array instances, each 64 bytes
- refactor `merkleizeInto()`:
  - rename `merkleizeBlocksBytes()` to avoid naming conflict
  - rename variables
- model block bytes which is 64 bytes instead of chunk bytes which is 32 bytes (1 block = 2 chunks)
- the consuming of this `merkleizeBlockArray()` method will come in the next PR

part of #378

### DIFF
--- a/packages/persistent-merkle-tree/src/hasher/util.ts
+++ b/packages/persistent-merkle-tree/src/hasher/util.ts
@@ -79,6 +79,10 @@ export function doMerkleizeBlocksBytes(
  * @param blockLimit number of blocks, should be <= blocks.length so that consumer can reuse memory
  * @param padFor is maxChunkCount, should be >= 2
  * @param blocks is unsafe because it's modified
+ * @param output the result is stored here
+ * @param offset the offset to store the result
+ * @param hashInto the hash function of each hasher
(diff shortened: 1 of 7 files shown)


packages/persistent-merkle-tree/src/hasher/util.ts
```diff
@@ -68,6 +74,115 @@ export function doMerkleizeInto(
...
+ * @param blocks is unsafe because it's modified
+ */
+export function doMerkleizeBlockArray(
+  blocks: Uint8Array[],
+  blockLimit: number,
+  padFor: number,
+  output: Uint8Array,
+  offset: number,
+  hashInto: HashIntoFn,
+  buffer: Uint8Array
```
wemeetagain, line 90: can you comment what `buffer` is used for?

wemeetagain (approved):

PR #423, opened 2024-11-21
feat: add tests workflow for bun and deno

nazarhussain:
**Motivation**

Make the packages compatible with Deno+Bun. 

**Description**

- Make unit tests compatible with Deno+Bun
- Make spec tests compatible with Deno+Bun

- Run all tests


packages/persistent-merkle-tree/src/hasher/util.ts
```diff
@@ -1,4 +1,4 @@
-import {byteArrayToHashObject, HashObject, hashObjectToByteArray} from "@chainsafe/as-sha256/lib/hashObject";
+import {byteArrayToHashObject, HashObject, hashObjectToByteArray} from "@chainsafe/as-sha256";
```
twoeths, line 1: 👍

twoeths:
while I see the improvement on the source code, like not to import `lib` from other modules, I don't see why do we have to pick Deno and make sure lodestar to be compliant to it? there are a lot of deno specific resources added in this PR

I guess we need to put this on hold until we have final approach on which javascript runtime that lodestar want to experiment https://github.com/ChainSafe/lodestar/issues/7237

nazarhussain:
> while I see the improvement on the source code, like not to import `lib` from other modules, I don't see why do we have to pick Deno and make sure lodestar to be compliant to it? there are a lot of deno specific resources added in this PR

Deno is needed/recommended to publish packages to jsr.io which is intention to test with deno first in this PR.

twoeths:
Does Bun support jsr.io? 

if yes I'd go with Bun as we'll most likely to go with it. Again, need to wait for the final decision on https://github.com/ChainSafe/lodestar/issues/7237

if both NodeJS + Bun don't support jsr.io, only Deno does then we need to see if it's a must to support jsr.io. I found it weird to target Bun/NodeJS while having Deno specific resources in the repo

nazarhussain:
> Does Bun support jsr.io?

jsr.io is a registry developed and maintained by Bun for typescript based packages. As it follow modern standards it's compatible with all runtimes. As it's mentioned on there website. 

> JSR modules can be used in Node.js, Deno, Bun, Cloudflare Workers, and more.

twoeths (approved):
lgtm

wemeetagain (approved):

PR #427, opened 2024-12-18
feat: non simd sha256 for incompatible systems

matthewkeil:
**Motivation**

An issue was posted about systems that do not support simd feature set.  Made library compatible with those systems.

See https://github.com/ChainSafe/lodestar/issues/7177 for more information.

**Description of Changes**
- `as-sha25`
  - make common.ts that holds non-simd functionality
  - make simd and non-simd entrance files
  - call common from both entrance files and implement simd only in simd version
  - modify codegen and package scripts to build both versions of the code
  - modify TS wrapper code to load correct version of bindings depending on simd presence
  - modify TS wrapper code to call correct methods depending on presence of simd
  - fix tests to exercise both sets of bindings fully


twoeths:
@matthewkeil I don't see we're using simd sha256 in lodestar, we haven't enabled batch hash yet

packages/as-sha256/assembly/index.ts
```diff
@@ -1,361 +1,13 @@
...
+    INPUT_LENGTH,
+    PARALLEL_FACTOR,
+    input,
+    output,
+    init,
+    update,
+    final,
+    digest,
+    digest64
+} from "./common";
```
wemeetagain, line 13: Add a non-simd implementation of `batchHash4UintArray64s` and `batchHash4HashObjectInputs` here (making the interface of simd and non-simd wasm module the same). Then the javascript layer won't need runtime checking and the only refactor there is the module reinitialization.

matthewkeil: I looked more deeply at this and there will still be conditional handling in the JS layer because for one of the methods because of how the data is copied.  

I was able to get `hashInto` and `batchHash4UintArray64s` converted relatively easily but in `batchHash4HashObjectInputs` the copy to the `inputUint32Array` interleaves the 4 hash objects and `digest64HashObject` requires the two hash object values `H0-H7`  to be contiguous as well as the two objects to be contiguous (see below).  So there is no easy way to proxy the pointer with offset easily.  The data will need to be handled differently for the two cases at the JS layer.  Do you have any ideas of how to easily do this?  I gave a couple thoughts but there will always need to be a conditional in TS land when loading `inputUint32Array`...

Screenshot 2025-01-09 at 8 41 43 PM

wemeetagain (approved):
LGTM!
looks like the repo also got linted?

PR #456, opened 2025-01-21
fix: ListCompositeTreeViewDU to getAllReadonly() without commit

twoeths:
**Motivation**
got `Must commit changes before reading all nodes` error in devnet-5 of electra https://github.com/ChainSafe/lodestar/pull/7375

**Description**

- getAllReadonly() without commit
  - we should be able to populate nodes array up until the old length
  - then get from the pending change, if not then get from the ViewDU of loaded `this.nodes`
- getAllReadonlyValues() still need the commit change because we want the nodes of pending changes. So although the name is similar, it's different from `getllReadonly()`

- unit test was added

cc @nflaig @wemeetagain

### DIFF
--- a/packages/ssz/src/viewDU/arrayComposite.ts
+++ b/packages/ssz/src/viewDU/arrayComposite.ts
@@ -145,14 +145,16 @@ export class ArrayCompositeTreeViewDU<
   }
 
   /**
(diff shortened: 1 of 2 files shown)


nflaig (approved):
This seems like the cleanest solution to the issue we have

wemeetagain (approved):

PR #459, opened 2025-01-23
feat!: use named exports to expose different hashers

nazarhussain:
**Motivation**

Allow users to easily choose the hasher. 

**Description**

- Use named exports for different hashers 
- Limit users to not load build directories directly

Run all tests

### DIFF
--- a/packages/persistent-merkle-tree/package.json
+++ b/packages/persistent-merkle-tree/package.json
@@ -3,8 +3,29 @@
   "version": "0.9.1",
   "description": "Merkle tree implemented as a persistent datastructure",
   "type": "module",
-  "main": "./lib/index.js",
-  "types": "./lib/index.d.ts",
+  "exports": {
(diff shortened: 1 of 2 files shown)


packages/persistent-merkle-tree/package.json
```diff
@@ -3,8 +3,33 @@
...
   "description": "Merkle tree implemented as a persistent datastructure",
   "type": "module",
-  "main": "./lib/index.js",
-  "types": "./lib/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./lib/index.js",
+      "types": "./lib/index.d.ts"
+    },
+    "./hashtree": {
```
nazarhussain, line 11: Other idea was to have `./hasher/hashtree` but I feel shorter path is better.

wemeetagain (approved):

PR #472, opened 2025-03-13
feat: get list composite by range

twoeths:
**Motivation**

- when the list length is big, application may want to get by chunk instead of get all of items because logic may break in the middle and using `getAllReadonly()` is not performant

see https://github.com/ChainSafe/lodestar/issues/7565

**Description**

- implement `getReadonlyByRange()` api
- unit test added


twoeths:
this fixes the issue on lodestar https://github.com/ChainSafe/lodestar/pull/7566#issuecomment-2721110865

wemeetagain (approved):

PR #476, opened 2025-03-14
chore: fix linting rule lint/style/noUselessElse

nazarhussain:
**Motivation**

**Description**

Closes #issue_number

### DIFF
--- a/packages/persistent-merkle-tree/src/node.ts
+++ b/packages/persistent-merkle-tree/src/node.ts
@@ -167,34 +167,29 @@ export class LeafNode extends Node {
       const h = getNodeH(this, hIndex);
       if (uintBytes === 1) {
         return 0xff & (h >> bitIndex);
-      } else {
-        return 0xffff & (h >> bitIndex);
       }
+      return 0xffff & (h >> bitIndex);
     }
 
     // number equals the h value
-    else if (uintBytes === 4) {
+    if (uintBytes === 4) {
       return getNodeH(this, hIndex) >>> 0;
     }
 
     // number spans 2 h values
-    else if (uintBytes === 8) {
+    if (uintBytes === 8) {
       const low = getNodeH(this, hIndex);
       const high = getNodeH(this, hIndex + 1);
-      if (high === 0) {
-        return low >>> 0;
-      } else if (high === -1 && low === -1 && clipInfinity) {
+      if (high === 0) return low >>> 0;
+      if (high === -1 && low === -1 && clipInfinity) {
         // Limit uint returns
         return Infinity;
-      } else {
-        return (low >>> 0) + (high >>> 0) * TWO_POWER_32;
       }
+      return (low >>> 0) + (high >>> 0) * TWO_POWER_32;
     }
 
     // Bigger uint can't be represented
-    else {
-      throw Error("uintBytes > 8");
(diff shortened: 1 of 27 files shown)


nazarhussain:
Deno tests will pass once this PR is merge. https://github.com/ChainSafe/ssz/pull/481

packages/persistent-merkle-tree/src/node.ts
```diff
@@ -251,7 +242,7 @@ export class LeafNode extends Node {
 
     // number spans 2 h values
     else if (uintBytes === 8) {
-      if (value === Infinity && clipInfinity) {
+      if (value === Number.POSITIVE_INFINITY && clipInfinity) {
```
matthewkeil, line 245: 😢 can we go back to `Infinity` here?  Same with the other "global" JS stuff that was namespaced to `Number` like `parseInt`.

packages/ssz/src/util/proof/treePartialToJsonPaths.ts
```diff
@@ -41,75 +41,67 @@ export function treePartialToJsonPaths(
       const childType = type.getPropertyType(jsonPathProp);
 
       // If this type merkleized fits in a single chunk then this LeafNode includes all data
-      if (childType.maxChunkCount === 1) {
-        return {type: TreeDataTypeCode.complete, jsonPathProps: [jsonPathProp]};
-      } else {
-        return {type: TreeDataTypeCode.witness};
-      }
+      if (childType.maxChunkCount === 1) return {type: TreeDataTypeCode.complete, jsonPathProps: [jsonPathProp]};
```
matthewkeil, line 44: Not having the brackets around conditional blocks is a recipe for bugs.  Even if there is only a single line I think we should keep them.  Only in some very rare cases where there are a bunch of very very short conditionals in sequence does this make sense


```ts
if (cond1) return "ans1";
if (cond2) return "ans2";
if (cond3) return "ans3";
if (cond4) return "ans4";
return "ans5";
```

packages/ssz/test/perf/misc/switch_if_cost.test.ts
```diff
@@ -5,16 +5,16 @@ describe("big if vs obj", () => {
...
+    if (n === 1) return 1;
+    if (n === 2) return 2;
+    if (n === 3) return 3;
+    if (n === 4) return 4;
+    if (n === 5) return 5;
+    if (n === 6) return 6;
+    if (n === 7) return 7;
+    if (n === 8) return 8;
+    if (n === 9) return 9;
+    throw Error("bad n");
```
matthewkeil, line 17: This is a perfect example of where no bracket conditionals are ok and even desirable.  But this type of code segment is about the only exception imho

matthewkeil (changes requested):
Couple small changes but otherwise looks good

packages/persistent-merkle-tree/src/node.ts
nazarhussain: Reverted that complete commit earlier, don't know how it reaches back in here. I will clean commit history for this branch.

matthewkeil (approved):
LGTM!  🚀 

Sorry it took so long to get back around to the final review!!! 😢

PR #480, opened 2025-03-27
feat: hashInto() and digest64Into() api

twoeths:
**Motivation**

- we have inefficient `executeHashComputations()` on Linux but have no idea how much overhead of this function

**Description**

- add more benchmarks to compare how much batch hash is better for each hasher
- the `hashInto()` is not comparable to `digest64()` because `digest64()` causes memory allocation, so I implemented `*Into()` apis to make a fair comparison
- for some scenarios in lodestar, it's better to use "*Into" apis


wemeetagain (approved):
lgtm

PR #481, opened 2025-04-07
chore: enforce vitest patch version to support unit testing for Deno

nazarhussain:
**Motivation**

We want to support testing of these packages on Deno and Bun. 

**Description**

The minor update to the `vitest` version is causing failure for Deno. Until resolve this issue we need to only support the patch version update. 

Run all tests


matthewkeil (approved):
LGTM!! 🚀

nflaig:
why do we wanna unit test on deno?

nazarhussain:
> why do we wanna unit test on deno?

For better safety while publishing to the JSR. As the JSR ecosystem is tightly developed around Deno support.

PR #484, opened 2025-05-02
chore: add support for ts extension import

nazarhussain:
**Motivation**

Support modern Typescript package registry. This is one not a full support for JSR but a step towards it. 

**Description**

- Update code to support Typescript direct import
 
Run all tests


nazarhussain:
That's the only point which might need attention for review. 

https://github.com/ChainSafe/ssz/pull/484/files#diff-1e8d3a9e1003ac505cd8c6fcf493b70a4c36fb6a5697f533f9827d202f6286abR230

packages/persistent-merkle-tree/src/node.ts
```diff
@@ -227,9 +227,11 @@ export class LeafNode extends Node {
       let h = getNodeH(this, hIndex);
       if (uintBytes === 1) {
         h &= ~(0xff << bitIndex);
+        // @ts-expect-error - This kind of expression is always truthy - Identified after TS 5.8 upgrade.
         h |= (0xff && value) << bitIndex;
```
wemeetagain, line 231: Ah, I suppose its found a bug. I think this is the intended
```suggestion
        h |= (0xff & value) << bitIndex;
```

nflaig (dismissed):
finally, the `.js` never made any sense

wemeetagain (approved):

PR #487, opened 2025-05-27
fix: check for AVX and return as-sha256 is missing

matthewkeil:
**Motivation**

A crash was reported by several parties running Lodestar on a VPS.  The problem was tracked down to no AVX support on the host which causes OffchainLabs/hashtree to crash.  The go binding do a similar check, at the bindings level, and revert to a non-assembly SHA256 lib.  This PR does the same.  Checks the hardware for the supported AVX types and returns `as-sha256` hasher from the `hashtree` hasher so there is no crash.

https://github.com/OffchainLabs/hashtree/blob/67979dccfc99b02b488ed8f5fddee9f09aea411e/src/hashtree.c#L53-L67


nflaig (approved):
seems good to me, leaving up to @wemeetagain or @twoeths for final approval

PR #489, opened 2025-05-29
fix: fix digest when input length = n * 64 - 1

### LINKED ISSUE #488: Incorrect digest()
**Describe the bug**

as found in [lodestar](https://github.com/ChainSafe/lodestar/issues/7879#issuecomment-2918128565), the `digest()` function returned incorrect result
this is a simplified unit test

```typescript
it('15364 digest test', function () {
    const MESSAGE_DOMAIN_VALID_SNAPPY = Buffer.from("01000000", "hex");
    const topic = "/eth2/82cb98a6/data_column_sidecar_3/ssz_snappy";
    const topicLength = new Uint8Array([topic.length, 0, 0, 0, 0, 0, 0, 0]);
    const data = Buffer.alloc(15364, 1);
    const input = Buffer.concat([MESSAGE_DOMAIN_VALID_SNAPPY, topicLength, Buffer.from(topic), data]);
    const output = Buffer.from(digest(input)).toString("hex");
    const expectedOutput = crypto.createHash("sha256").update(input).digest().toString("hex");
    expect(output).to.be.equal(expectedOutput, "15364 digest test failed");
  });
```

some context:
- it does not happen if we don't concat `MESSAGE_DOMAIN_VALID_SNAPPY` (4 bytes)
- it does not happen if we don't concat topicLength (8 bytes)
- it does not happen if we don't concat `Buffer.from(topic)`
- it does not happen with shorter data length (<= 15299)

**Expected behavior**

- it should always return the same result to NodeJS
- add more unit tests to compare to NodeJS's sha256

**Steps to Reproduce**
- run the above unit tests, add more if needed

wemeetagain:
**Motivation**

- resolve #488 

**Description**

- simplify (and correct) the padding logic (see https://www.rfc-editor.org/rfc/rfc6234#section-4)

### DIFF
--- a/packages/as-sha256/assembly/common.ts
+++ b/packages/as-sha256/assembly/common.ts
@@ -254,20 +254,15 @@ export function update(dataPtr: usize, dataLength: i32): void {
 }
 
 export function final(outPtr: usize): void {
-  // one additional round of hashes required
-  // because padding will not fit
-  if ((bytesHashed & 63) < 63) {
-    store8(mPtr, mLength, 0x80);
-    mLength++;
-  }
-  if ((bytesHashed & 63) >= 56) {
+  // add padding bit
+  store8(mPtr, mLength, 0x80);
+  mLength++;
+
+  // one additional round of hashes required if the input length can't fit in the current block
+  if (mLength > 56) {
     fill(mPtr + mLength, 0, 64 - mLength);
     hashBlocks(wPtr, mPtr, 1);
     mLength = 0;
-  }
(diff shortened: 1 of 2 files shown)


twoeths (approved):
the fix makes sense to me, this EIP helped me a lot to understand it and sha256 in general https://eips.ethereum.org/EIPS/eip-7797

I hope this is helpful for someone

matthewkeil (approved):
Wowwy zowwy you are a 🥷!

PR #492, opened 2025-06-18
fix: re-enable noParamaterReAssign and fix linting

ChefBingbong:
**Motivation**
This PR takes care of a pending task, which is to re-enable the biome `noParamaterReAssign` style lint and to fix all instances in the codebase where this rule is breached

changes
1. Re-enabled biome's `noParamaterReAssign`
2. fixed all places where this rule was breached,
3. ensured the lint command in root runs with no warnings
4. ensured all test suites remained unaffected by the relevant code changes


nazarhussain (changes requested):
The changes related to primitives should be removed. All values of `number`, `bigint` or any other primitive is passed by value, so no issue if these are reassigned in local function scope. 

We should focus the changes only object or complex types where accidental parameter change have any side effect.

PR #499, opened 2025-08-13
fix: optimize the leaf node creation

nazarhussain:
**Motivation**

Improve the performance of lead node creation during the packing values. 

**Description**

- Initialize primitive values instead of `LeafNode` object
- Create the `LeftNode` object once with those values. 

Based on the work started in this PR https://github.com/ChainSafe/ssz/pull/493

Run all tests.


twoeths (commented):
would be great to know how does this PR impacts the performance through benchmarks

PR #504, opened 2025-11-22
chore: remove deno test

wemeetagain:
**Motivation**

- CI broken and annoying


matthewkeil:
LGTM! Can turn back on after release goes out

PR #512, opened 2026-03-17
fix: allow array-like input for VectorBasicType views

nflaig:
This allows to pass `Uint32Array` as value


wemeetagain (approved):
lgtm. essentially just a relaxation of the param type from `thing[]` to `ArrayLike<thing>`

PR #517, opened 2026-04-21
chore: bump node to v24

nflaig:
**Motivation**

The publish workflow fails. Possibly because npm needs to be > 11.5.1 for OIDC publishing.   Updating node in publish workflow to try


matthewkeil (approved):
LGTM 🚀

PR #525, opened 2026-04-21
feat: allow array-like input for VectorCompositeType views

nflaig:
Continuation of https://github.com/ChainSafe/ssz/pull/512, to apply the same change `VectorCompositeType` views

When the element is a `VectorBasicType` or `ListBasicType`, accept any `ArrayLike` (eg. `Uint32Array`) in place of `number[]`.


wemeetagain (approved):
lgtm

PR #536, opened 2026-07-14
fix: avoid stack overflow materializing large progressive lists

### LINKED ISSUE #535: ProgressiveList: "Maximum call stack size exceeded" materializing large lists (spread-push in getNodesAtProgressiveDepth)
## Summary

`getNodesAtProgressiveDepth` (`packages/ssz/src/type/progressive.ts:159`) materializes its result with `nodes.push(...getNodesAtDepth(...))`. When a `ProgressiveList` is large enough that a single progressive subtree holds more than V8's spread/apply argument limit (~65 535), the spread expands to that many call arguments and throws `RangeError: Maximum call stack size exceeded`. Any `forEachValue` / `getAllReadonly` / `populateAllNodes` over a sufficiently large progressive list crashes.

Present in `ssz-v1.6.1` (latest) and on `main`.

## Where

```ts
// packages/ssz/src/type/progressive.ts
159:    nodes.push(...getNodesAtDepth(node.left, subtreeDepth, 0, subtreeNodeCount));
```

With `BASE_CHUNK_COUNT = 1`, `SCALING_FACTOR = 4`, `progressiveSubtreeDepth(i) = 2*i`, the per-subtree node counts are `1, 4, 16, 64, … 16384, 65536, 262144, …`. `getNodesAtDepth(..., subtreeNodeCount)` returns up to `subtreeNodeCount` nodes, so once a subtree reaches 65 536 the `push(...)` spread overflows. In practice any progressive list with length ≳ 87 381 (where the 65 536-node subtree is first reached) hits it deterministically.

## Real stack trace

Observed on Lodestar `v1.44.0` on the `glamsterdam-devnet-7` Gloas/ePBS devnet:

```
RangeError: Maximum call stack size exceeded
    at getNodesAtProgressiveDepth (@chainsafe/ssz/src/type/progressive.ts:159:11)
    at ProgressiveListCompositeTreeViewDU.populateAllNodes (@chainsafe/ssz/src/type/progressiveList.ts:1013:20)
    at ProgressiveListCompositeTreeViewDU.forEachValue (@chainsafe/ssz/src/type/progressiveList.ts:923:10)
    at getEffectiveBalanceIncrementsZeroInactive (lodestar/packages/state-transition/src/util/balance.ts:67)
    at CheckpointBalancesCache.processState (lodestar/packages/beacon-node/src/chain/balancesCache.ts:38)
    at BeaconChain.importBlock (lodestar/packages/beacon-node/src/chain/blocks/importBlock.ts:114)
```

## Repro / impact

On that devnet the genesis preloads ~500 000 validators (registry length ≈ 503 000) and `state.validators` is materialized via the `ProgressiveList` view. The first block that walks the full registry — `getEffectiveBalanceIncrementsZeroInactive` doing `state.validators.forEachValue(...)` → `populateAllNodes()` → `getNodesAtProgressiveDepth(chunksNode, 503000)` — hits a largest subtree of `262 144` nodes and overflows the spread. Every subsequent block fails identically, so all Lodestar nodes are permanently stuck at that boundary. A smaller net (~8k validators → largest subtree 4096) is unaffected. Any mainnet-scale progressive list (>1M items) would hit this deterministically.

## Suggested fix

Append without expanding into call arguments. Minimal:

```ts
const subNodes = getNodesAtDepth(node.left, subtreeDepth, 0, subtreeNodeCount);
for (let i = 0; i < subNodes.length; i++) nodes.push(subNodes[i]);
```

or have `getNodesAtDepth` write directly into a caller-provided target to avoid the intermediate allocation. Worth auditing sibling spreads over view-derived arrays (e.g. `progressiveList.ts:521` `gindices.push(...tree_getLeafGindices(...))`) and the `getAll*` paths for the same footgun.

## Versions

`@chainsafe/ssz` `1.6.1` and `main`.

lodekeeper:
## Problem

`getNodesAtProgressiveDepth` materialized its result with `nodes.push(...getNodesAtDepth(...))`, spreading an entire progressive subtree as function arguments. Progressive subtrees grow geometrically (1, 4, 16, … 65536, 262144, …), so for a large list a single subtree's node array is spread into hundreds of thousands of call arguments and V8 throws `RangeError: Maximum call stack size exceeded`.

This is a deterministic crash at scale. On the `glamsterdam-devnet-7` Gloas/ePBS devnet, Lodestar nodes with a ~500k-validator registry (largest subtree = 262144) permanently stalled at the fork boundary — every block that walked `state.validators` via `forEachValue` → `populateAllNodes` → `getNodesAtProgressiveDepth` failed with:

```
RangeError: Maximum call stack size exceeded
    at getNodesAtProgressiveDepth (@chainsafe/ssz/src/type/progressive.ts:159)
    at ProgressiveListCompositeTreeViewDU.populateAllNodes (progressiveList.ts:1013)
    at ProgressiveListCompositeTreeViewDU.forEachValue (progressiveList.ts:923)
    at getEffectiveBalanceIncrementsZeroInactive (lodestar state-transition)
```

Fixes #535.

## Fix

Pre-allocate the result (`new Array<Node>(count)`) and fill by offset instead of spreading. `getNodesAtDepth` returns exactly `count` nodes, so the per-subtree counts sum to `count`. Output ordering and node identity are unchanged, and this avoids the intermediate array-growth from repeated `push`.

## Test

Adds a `getNodesAtProgressiveDepth` test that builds a list filling the 9th subtree completely (262144 nodes — the exact shape seen in production) and asserts every node is returned in order across the subtree boundary.

Note: the overflow threshold is stack-size dependent (≈125k spread args on Node's default main-thread stack; higher under vitest's worker), so the test primarily locks in correctness of large-list materialization at production scale — the unbounded spread is removed entirely by this change.

## Verification

- `pnpm --filter @chainsafe/ssz exec vitest run test/unit/byType/progressive/valid.test.ts` → **20 passed**
- Reverting only the source change reproduces `RangeError: Maximum call stack size exceeded` on Node's default stack at the affected sizes
- `biome check` and `tsc --noEmit` clean for the changed files

### DIFF
--- a/packages/ssz/src/type/progressive.ts
+++ b/packages/ssz/src/type/progressive.ts
@@ -144,9 +144,15 @@ export function progressiveSubtreeFillToContents(nodes: Node[]): Node {
 }
 
 export function getNodesAtProgressiveDepth(rootNode: Node, count: number): Node[] {
-  const nodes: Node[] = [];
+  // Pre-allocate the exact result size and fill by offset. Do NOT use
+  // `nodes.push(...getNodesAtDepth(...))`: a single progressive subtree can hold up to 2^(2*subtreeIndex)
+  // chunks, so for large lists the spread expands to hundreds of thousands of call arguments and throws
+  // `RangeError: Maximum call stack size exceeded`. `getNodesAtDepth` returns exactly `count` nodes, so the
+  // per-subtree counts sum to `count`.
+  const nodes = new Array<Node>(count);
   let node = rootNode;
   let remaining = count;
+  let offset = 0;
   let subtreeLength = BASE_CHUNK_COUNT;
 
   for (let subtreeIndex = 0; remaining > 0; subtreeIndex++) {
@@ -156,7 +162,11 @@ export function getNodesAtProgressiveDepth(rootNode: Node, count: number): Node[
 
     const subtreeNodeCount = Math.min(subtreeLength, remaining);
     const subtreeDepth = progressiveSubtreeDepth(subtreeIndex);
-    nodes.push(...getNodesAtDepth(node.left, subtreeDepth, 0, subtreeNodeCount));
+    const subtreeNodes = getNodesAtDepth(node.left, subtreeDepth, 0, subtreeNodeCount);
+    for (let i = 0; i < subtreeNodes.length; i++) {
+      nodes[offset + i] = subtreeNodes[i];
+    }
+    offset += subtreeNodeCount;
     node = node.right;
     remaining -= subtreeNodeCount;
     subtreeLength *= SCALING_FACTOR;
--- a/packages/ssz/src/type/progressive.ts
+++ b/packages/ssz/src/type/progressive.ts
@@ -144,11 +144,7 @@ export function progressiveSubtreeFillToContents(nodes: Node[]): Node {
 }
 
 export function getNodesAtProgressiveDepth(rootNode: Node, count: number): Node[] {
-  // Pre-allocate the exact result size and fill by offset. Do NOT use
-  // `nodes.push(...getNodesAtDepth(...))`: a single progressive subtree can hold up to 2^(2*subtreeIndex)
-  // chunks, so for large lists the spread expands to hundreds of thousands of call arguments and throws
-  // `RangeError: Maximum call stack size exceeded`. `getNodesAtDepth` returns exactly `count` nodes, so the
-  // per-subtree counts sum to `count`.
+  // Fill by offset: spreading a large subtree into `push(...)` exceeds V8's argument limit.
   const nodes = new Array<Node>(count);
   let node = rootNode;
   let remaining = count;
--- a/packages/ssz/src/type/progressive.ts
+++ b/packages/ssz/src/type/progressive.ts
@@ -144,7 +144,6 @@ export function progressiveSubtreeFillToContents(nodes: Node[]): Node {
 }
 
 export function getNodesAtProgressiveDepth(rootNode: Node, count: number): Node[] {
-  // Push nodes individually; `nodes.push(...subtreeNodes)` would spread a large subtree past V8's argument limit.
   const nodes: Node[] = [];
   let node = rootNode;
   let remaining = count;
--- a/packages/ssz/src/type/progressive.ts
+++ b/packages/ssz/src/type/progressive.ts
@@ -144,11 +144,10 @@ export function progressiveSubtreeFillToContents(nodes: Node[]): Node {
 }
 
 export function getNodesAtProgressiveDepth(rootNode: Node, count: number): Node[] {
-  // Fill by offset: spreading a large subtree into `push(...)` exceeds V8's argument limit.
-  const nodes = new Array<Node>(count);
+  // Push nodes individually; `nodes.push(...subtreeNodes)` would spread a large subtree past V8's argument limit.
+  const nodes: Node[] = [];
   let node = rootNode;
   let remaining = count;
-  let offset = 0;
   let subtreeLength = BASE_CHUNK_COUNT;
 
   for (let subtreeIndex = 0; remaining > 0; subtreeIndex++) {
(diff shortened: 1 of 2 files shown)


ensi321:
https://github.com/ChainSafe/ssz/blob/ddfcaf56183ff6aa9a038ad57447b971bc29c8e0/packages/ssz/src/type/progressiveList.ts#L521
Do we need to fix this too? @lodekeeper

packages/ssz/src/type/progressive.ts
```diff
@@ -144,9 +144,15 @@ export function progressiveSubtreeFillToContents(nodes: Node[]): Node {
 }
 
 export function getNodesAtProgressiveDepth(rootNode: Node, count: number): Node[] {
-  const nodes: Node[] = [];
+  // Pre-allocate the exact result size and fill by offset. Do NOT use
+  // `nodes.push(...getNodesAtDepth(...))`: a single progressive subtree can hold up to 2^(2*subtreeIndex)
+  // chunks, so for large lists the spread expands to hundreds of thousands of call arguments and throws
+  // `RangeError: Maximum call stack size exceeded`. `getNodesAtDepth` returns exactly `count` nodes, so the
+  // per-subtree counts sum to `count`.
+  const nodes = new Array<Node>(count);
```
wemeetagain: This verbose comment is not helpful.

lodekeeper: Shortened to a one-liner in b0e5264:

```ts
// Fill by offset: spreading a large subtree into `push(...)` exceeds V8's argument limit.
```

Kept a short note on the *why* so the offset-fill doesn't get "simplified" back into a spread that reintroduces the `RangeError`.

lodekeeper:
Good question — looked at it closely. Short answer: **not the same failure mode; no change needed here.**

The bug this PR fixes is a *single* spread whose size scales with **list length**: `getNodesAtProgressiveDepth` did `nodes.push(...getNodesAtDepth(node.left, subtreeDepth, 0, subtreeNodeCount))`, and one progressive subtree holds up to `2^(2*subtreeIndex)` chunks — so for a large list a single call spreads hundreds of thousands of args → `RangeError`.

`progressiveList.ts` L521 (and L518) is different: the spread is **per element**, inside the `for (i < length)` loop. Each `gindices.push(...this.elementType.tree_getLeafGindices(...))` spreads only *one element's* leaf gindices, bounded by the **element type's** leaf count — not the list length. The list length drives the *number* of pushes, not the size of any single spread. For realistic cases (basic → no spread; container / fixed-size composite → a small fixed leaf count) this is nowhere near V8's argument limit.

The only way one of these per-element spreads gets large is a pathologically large *element type* (e.g. a nested `List`/`Vector`/`ProgressiveList` with tens of thousands of leaves in a single element). But that isn't specific to `ProgressiveList` — it's the same `gindices.push(...<type>.tree_getLeafGindices(...))` pattern used across the board: `array.ts` (List/Vector composite), `container.ts`, `union.ts`, `profile.ts`, `stableContainer.ts`, `progressiveContainer.ts`. So if we want to harden against giant nested element types, it should be one consistent change across all of them, not a one-off here — and it's orthogonal to this list-length stack-overflow fix.

I'd keep this PR focused and open a separate issue for the generic pattern if we think it's worth doing. Happy to take that follow-up if you agree 👍

ensi321 (commented):
lgtm. wait for @wemeetagain 's approval

packages/ssz/src/type/progressive.ts
```diff
@@ -144,9 +144,11 @@ export function progressiveSubtreeFillToContents(nodes: Node[]): Node {
 }
 
 export function getNodesAtProgressiveDepth(rootNode: Node, count: number): Node[] {
-  const nodes: Node[] = [];
+  // Fill by offset: spreading a large subtree into `push(...)` exceeds V8's argument limit.
+  const nodes = new Array<Node>(count);
```
wemeetagain: I would prefer to see empty array initialization here and nodes.push(singleItem) below.

lodekeeper: Done in 57abfa6 — empty-array init + `nodes.push(subtreeNodes[i])`, dropped the `offset` variable and the preallocated `new Array<Node>(count)`.

Kept a one-line note pointing at the anti-pattern (`nodes.push(...subtreeNodes)`) so the loop doesn't get "optimized" back into the large spread that caused #535. Progressive unit tests green (20/20), including the large-list regression.

packages/ssz/src/type/progressive.ts
```diff
@@ -144,6 +144,7 @@ export function progressiveSubtreeFillToContents(nodes: Node[]): Node {
 }
 
 export function getNodesAtProgressiveDepth(rootNode: Node, count: number): Node[] {
+  // Push nodes individually; `nodes.push(...subtreeNodes)` would spread a large subtree past V8's argument limit.
```
wemeetagain:
```suggestion
```

wemeetagain (approved):

packages/ssz/src/type/progressive.ts
lodekeeper: Applied — the flagged comment line was dropped as part of the cleanup in this PR; confirmed it's gone from `packages/ssz/src/type/progressive.ts` on the merged `main`.

PR #537, opened 2026-07-14
fix: avoid stack overflow in proof gindices for large lists/fields

lodekeeper:
Follow-up to #536, addressing #9656.

## Problem

#536 fixed the `getNodesAtProgressiveDepth` crash that actually stalled Lodestar at the Gloas fork. The **same `x.push(...y)` footgun** exists throughout the **proof-gindices** path: nearly every composite type accumulates child leaf gindices with `parent.push(...child.tree_getLeafGindices(...))`.

Spreading an array into a call passes each element as a separate argument, and V8 caps call arguments by stack size (~125k on Node's default stack). The `container`/`stableContainer`/`profile`/`array` sites spread an **entire field/element's** leaf gindices, so a proof that descends into a large collection field (e.g. a mainnet-scale `validators` list) throws `RangeError: Maximum call stack size exceeded` the same way. It's latent today — normal block import doesn't build these proofs — but reachable on large states.

## Change

- Add `pushAll(target, source)` (`packages/ssz/src/util/array.ts`): appends in a loop, no spread.
- Replace all **19** `x.push(...y)` spreads across `array`, `composite`, `container`, `compatibleUnion`, `union`, `profile`, `progressiveContainer`, `stableContainer`, `progressiveList` with `pushAll`. Behaviour (order, output) is unchanged.
- Add `pnpm lint:no-push-spread` + a CI step to ban the pattern from `packages/ssz/src` so it can't come back.

Note on the lint rule: biome 1.9.4 has no custom-rule / GritQL-plugin support (that lands in biome 2.x), so the guard wraps biome's experimental GritQL `search` command (`scripts/no-push-spread.mjs`). It can become a native plugin when biome is upgraded.

## Verification

- `pnpm test:unit` (ssz): **1783 passed / 3 skipped**
- Proof tests (`proof`, `proofs`, `compatibleUnion`, `progressive`): green
- New `pushAll` unit tests incl. a 300k-element case that a spread would overflow
- `pnpm lint:no-push-spread`: passes clean; verified it fails on a planted `a.push(...b)`
- `biome check` + `tsc --noEmit` clean on changed files


packages/ssz/src/type/array.ts
```diff
@@ -86,11 +87,11 @@ export abstract class ArrayType<ElementType extends Type<unknown>, TV, TVDU> ext
           gindex < endGindex;
           gindex++, extendedGindex++
         ) {
-          gindices.push(...this.elementType.tree_getLeafGindices(extendedGindex, getNode(rootNode, gindex)));
+          pushAll(gindices, this.elementType.tree_getLeafGindices(extendedGindex, getNode(rootNode, gindex)));
```
wemeetagain: Please remove the pushAll helper and just use vanilla push.

lodekeeper: Before I sweep this — "vanilla push" reads two ways, and they treat the `no-push-spread` guard oppositely:

- **(A)** inline `for (const g of …tree_getLeafGindices(…)) gindices.push(g)` — keeps the large-list safety and the lint stays meaningful.
- **(B)** plain `gindices.push(…tree_getLeafGindices(…))` spread — all 19 sites here are per-element/per-field, so the spread is bounded by the element/field type's own leaf count (not the list length), i.e. safe for every realistic type. This would mean dropping the `no-push-spread` guard, since we'd be intentionally allowing bounded spreads.

(The only genuinely *unbounded* spread was `getNodesAtProgressiveDepth`, which is already a loop from #536.)

I lean **(A)** so #537 still hardens these paths without the helper abstraction, but **(B)** is tidier if you consider the guard overkill for bounded sites. Which do you want?

packages/ssz/src/util/array.ts
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Prefer this over `target.push(...source)`: spreading an array into a call passes each element as a
+ * separate argument, and V8 caps call arguments by stack size (~125k on Node's default stack). For large
+ * `source` arrays the spread throws `RangeError: Maximum call stack size exceeded`. This matters in the
+ * proof/gindices paths, where a single subtree or collection field can produce hundreds of thousands of
+ * elements (see chainsafe/ssz#535).
+ */
+export function pushAll<T>(target: T[], source: ArrayLike<T>): void {
+  for (let i = 0; i < source.length; i++) {
+    target.push(source[i]);
```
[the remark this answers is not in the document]
lodekeeper: Valid — real latent footgun: the loop re-reads `source.length` each iteration, so a self-aliasing `pushAll(arr, arr)` (the `arr.push(...arr)` replacement case) would never terminate. No current call site aliases `target`, but per @wemeetagain's review the `pushAll` helper is being removed in favor of inline vanilla-push loops, so this goes away with it. Thanks 👍

packages/ssz/src/type/array.ts
wemeetagain: yes A

lodekeeper: Done — removed the `pushAll` helper and its test file, replaced all 9 call sites with inline `for...of` loops (option A). Lint + 1780 unit tests green.

wemeetagain (approved):

PR #539, opened 2026-07-14
chore: reduce test workflow critical path

wemeetagain:
## What changed

- Run feature branches only through `pull_request`; keep `push` runs for `master`
- Cancel superseded runs for the same PR
- Keep fast build, type, lint, unit, and browser checks on Node 22 and 24
- Run the expensive spec corpus once on Node 24
- Split spec tests into four parallel shards: general, phase0, altair, and bellatrix
- Preserve the required `Tests Node (22)` and `Tests Node (24)` check names as aggregate gates over both Node checks and every spec shard
- Pin the GitHub actions used by this workflow to the current versions already used by Lodestar

## Why

A recent ssz PR run took 28m38s. About 27 minutes were spent running seven spec commands serially in each Node matrix job. Lodestar keeps its critical path shorter by avoiding duplicate feature-branch push runs, cancelling stale PR runs, testing one primary Node version for expensive suites, and separating independent test families into jobs.

## Measured result

The first end-to-end run completed successfully in **8m08s**, down from **28m38s**:

- Node 22 checks: 1m00s
- Node 24 checks: 58s
- Bun: 38s
- phase0: 6m56s
- general: 7m10s
- altair: 7m21s
- bellatrix: 8m01s

That is about **72% less wall-clock time** and a **3.5× speedup** while retaining every existing test suite.

## Validation

- [End-to-end GitHub Actions run](https://github.com/ChainSafe/ssz/actions/runs/29364111877)
- `actionlint v1.7.12 .github/workflows/test.yml`
- YAML parser validation
- `git diff --check`

### DIFF
--- a/packages/as-sha256/test/perf/index.test.ts
+++ b/packages/as-sha256/test/perf/index.test.ts
@@ -1,4 +1,4 @@
-import {bench, describe, setBenchOpts} from "@chainsafe/benchmark";
+import {bench, describe} from "@chainsafe/benchmark";
 import {
   byteArrayToHashObject,
   digest,
@@ -13,10 +13,6 @@ import {
 //     ✓ digest64 50023 times                                                27.30382 ops/s    36.62491 ms/op   x1.003       1625 runs   60.0 s
 //     ✓ digest 50023 times                                                  27.31207 ops/s    36.61385 ms/op   x0.999       1624 runs   60.0 s
 describe("digestTwoHashObjects vs digest64 vs digest", () => {
-  setBenchOpts({
-    minMs: 60000,
-  });
-
   const input = Buffer.from("gajindergajindergajindergajindergajindergajindergajindergajinder", "utf8");
   const input1 = "gajindergajindergajindergajinder";
   const input2 = "gajindergajindergajindergajinder";
--- a/.benchrc.yaml
+++ b/.benchrc.yaml
@@ -2,6 +2,11 @@
 threshold: 3
 maxMs: 60000
 minRuns: 10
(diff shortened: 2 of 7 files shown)


packages/as-sha256/test/perf/index.test.ts
```diff
@@ -13,10 +13,6 @@ import {
 //     ✓ digest64 50023 times                                                27.30382 ops/s    36.62491 ms/op   x1.003       1625 runs   60.0 s
 //     ✓ digest 50023 times                                                  27.31207 ops/s    36.61385 ms/op   x0.999       1624 runs   60.0 s
 describe("digestTwoHashObjects vs digest64 vs digest", () => {
-  setBenchOpts({
-    minMs: 60000,
```
spiral-ladder, line 17: removing this to unblock for now seems fine but we prob should look at lowering this instead of removing completely, fable says:

> Fast micro-benches (as-sha256): minMs: 2000–5000, keep default convergence. That's thousands-to-millions of runs — plenty of samples.
Heavy cases (beaconState, validators): minMs: 5000, maxMs: 30000. Here each run is long, so what matters more is minRuns (keep ~10) and cheap setup, not wall time.

spiral-ladder (approved):
lgtm w ocmment

packages/as-sha256/test/perf/index.test.ts
wemeetagain: Yes tbh we should probably go thru each benchmark and disable ones which are no longer useful

PR #541, opened 2026-07-15
chore(ci): parallelize bench

spiral-ladder:
attempt at speeding up ci similar to #539

i kinda thnk splitting up the jobs is useful even as a signal as to which benchmark is slowest, seems like pmt is

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -15,9 +15,16 @@ on:
 
 jobs:
   benchmark:
+    name: benchmark (${{ matrix.package }})
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -41,5 +52,68 @@ jobs:
...
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          pattern: bench-output-*
+          merge-multiple: true
+
+      - name: Post aggregated comment
+        uses: actions/github-script@v7
+        with:
+          script: |
```
spiral-ladder, line 85: this is really ugly though but reasonably short (~30 LOC)? i dk how i feel about this

spiral-ladder:
the comment is botched, converting back to draft first

PR #543, opened 2026-08-14
fix: check first offset in readVariableOffsetsArrayComposite

twoeths:
**Motivation**

- we don't want to allocate too much memory if someone feed an invalid first offset in array composite ssz bytes

**Description**

- bound check it against the size for 1st offset
- we already check it from the 2nd offset


nflaig (approved):
seems good to me, @wemeetagain should probably give this a review

wemeetagain (approved):

